PR #1, opened 2017-09-18
Initial bootstrap

NaomiFPassarelli:
- Create initial bootstrap with yeoman basic structure and packages needed
- Added possibility to choose optional packages (carrousel, moment, etc)


generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
+{
+  "name": "<%= projectNameSlugified %>",
+  "title": "<%= projectName %>",
+  "description": "<%= projectDescription %>",
+  "version": "0.0.0",
```
SKOLZ: is it correct to have 0.0.0? I think we should set it to at least 0.1.0

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+  "repository": {
+    "type": "git",
+    "url": "<%= repoUrl %>"
+  },
+  "dependencies": {
+    <% if (includeAphrodite && !includeRadium) {%>
```
SKOLZ: remove aphrodite

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+      "aphrodite": "^1.2.3",<%
+    }
+
+    if (!includeAphrodite && includeRadium) {%>
```
SKOLZ: remove radium

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+    "radium": "^0.19.4",<%
+    }
+
+    if (includeInext) {%>
+    "i18next": "^9.0.0",<%
+    }
+
+    if (includeMobileDetect) {%>
```
SKOLZ: I don't think this is necesary, do we have an example of a project where we are using it?

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+
+    if (includeNukaCarousel) {%>
+    "nuka-carousel": "^2.3.0",<%
+    }
+
+    if (includeNumeral) {%>
+    "numeral": "^2.0.6",<%
+    }
+
+    if (includePostcss) {%>
```
SKOLZ: we should add this by default, I would also add with it the autoprefixer(https://github.com/postcss/autoprefixer) and the sprites(https://github.com/2createStudio/postcss-sprites) plugins

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+
+    if (includeReactAlert) {%>
+    "react-alert": "^2.3.0",<%
+    }
+
+    if (includeReactModal) {%>
+    "react-modal": "^2.3.2",<%
+    }
+
+    if (includeReactResponsive) {%>
```
SKOLZ: I would remove this from the selection. This is not used very often, I would use that library only  when we need it or directly define a utils function to determine a phone width and apply decitions based on that

SKOLZ (changes requested):

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+
+    if (includeRecharts) {%>
+    "recharts": "^1.0.0-alpha.4",<%
+    }
+
+    if (includeReactGoogleMaps) {%>
+    "react-google-maps": "^7.3.0",<%
+    }
+
+    if (includeReselect) {%>
```
SKOLZ: include this by default

generators/app/templates/public/index.html
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="initial-scale=1">
+  <meta charset="utf-8">
+  
+  <!--
```
SKOLZ: remove this comment

generators/app/templates/public/index.html
```diff
@@ -0,0 +1,32 @@
...
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+  <title> <% this.projectName %> </title>
+</head>
+
+<body>
+  <div id="root"></div>
+  <!--
```
SKOLZ: remove this comment

generators/app/templates/_package.json
mcallegari10: I agree, I'd use 1.0.0 instead

generators/app/templates/public/index.html
```diff
@@ -0,0 +1,32 @@
...
+      This HTML file is a template.
+      If you open it directly in the browser, you will see an empty page.
+      You can add webfonts, meta tags, or analytics to this file.
+      The build step will place the bundled scripts into the <body> tag.
+      To begin the development, run `npm start`.
+      To create a production bundle, use `npm run build`.
+    -->
```
mcallegari10: I'd delete this comment. Otherwise, it makes this a copy paste of create-react-app and we don't want that :smile:

generators/app/templates/_package.json
SKOLZ: add this by default

generators/app/templates/_package.json
```diff
@@ -0,0 +1,117 @@
...
+    "history": "^4.7.2",
+    "lodash": "^4.17.4",
+    "prop-types": "^15.5.10",
+    "react": "^15.6.1",
+    "react-addons-perf": "^15.4.2",
+    "react-dom": "^15.6.1",
+    "react-redux": "^5.0.6",
+    "react-router": "^4.2.0",
+    "react-router-dom": "^4.2.2",
+    "react-router-redux": "^5.0.0-alpha.6",
```
SKOLZ: I would add this one if we are using redux or the react router 4 if we are not.

mcallegari10 (changes requested):

generators/app/templates/public/index.html
SKOLZ: remove comment

generators/app/templates/public/index.html
SKOLZ: remove comment

generators/app/templates/src/app/components/Button/index.js
```diff
@@ -0,0 +1,71 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import PropTypes from "prop-types";
+import Radium from "radium";
```
SKOLZ: no radium, use css-modules instead

SKOLZ (changes requested):

generators/app/templates/eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+  parser: 'babel-eslint',
+  parserOptions: {
+    ecmaFeatures: {
+      experimentalObjectRestSpread: true,
+      jsx: true
+    },
+    sourceType: 'module'
+  },
+  plugins: ['react', 'prettier', 'flowtype'],
```
SKOLZ: we won't be using typed js

generators/app/templates/eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+  extends: [
+    'eslint:recommended',
+    'airbnb',
+    'plugin:react/recommended',
+    'prettier',
+    'prettier/flowtype',
```
SKOLZ: same

generators/app/templates/src/app/components/Checkbox/index.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import Radium from "radium";
```
SKOLZ: change radium to css modules here too please

generators/app/templates/src/app/components/Label/index.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import Radium from "radium";
```
SKOLZ: change radium to css modules here too please

generators/app/templates/src/app/components/RadioButton/index.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import Radium from "radium";
```
SKOLZ: change radium to css modules here too please

SKOLZ (changes requested):

pabloferro (dismissed):
Fix the package version and the debugger and 👍

generators/app/index.js
```diff
@@ -0,0 +1,502 @@
...
+            : `${val} is not a valid description for a project. Please use a valid description (alphanumeric).`
+      },
+      {
+        type: "input",
+        name: "repoUrl",
+        message: "What is the git repo url for this project?",
+        store: true
+      },
+      {
+        type: "confirm",
```
sbalay, line 68: I'd remove all these confirms and decide which libraries we want to always include and which libraries we don't.

When you kickoff a project you probably don't know if you will need recharts, carousel, mobileDetect, etc. Forcing a decision over something you don't know is more a problem than a solution.

generators/app/index.js
```diff
@@ -0,0 +1,502 @@
+"use strict";
```
sbalay: this file is huge

SKOLZ (dismissed):

mcallegari10 (dismissed):

generators/app/index.js
```diff
@@ -0,0 +1,502 @@
...
+      }
+    ];
+
+    return this.prompt(prompts).then(
+      function(answers) {
+        for (var key in answers) {
+          this[key] = answers[key];
+        }
+
+        function hasFeature(feat, prop) {
```
sbalay: What's the goal of this function?

generators/app/index.js
```diff
@@ -0,0 +1,502 @@
...
+
+    this.fs.copyTpl(
+      this.templatePath("src/index.js"),
+      this.destinationPath("src/index.js"),
+      {}
+    );
+
+    mkdirp(this.destinationPath("src/app/assets/"));
+
+    this.fs.copyTpl(
```
sbalay: Are all these files templates? Use `this.fs.copy` if they are not

generators/app/index.js
```diff
@@ -0,0 +1,502 @@
...
+      {}
+    );
+    this.fs.copyTpl(
+      this.templatePath("eslintrc.js"),
+      this.destinationPath(".eslintrc.js"),
+      {}
+    );
+
+    this.fs.copyTpl(
+      this.templatePath("public/index.html"),
```
sbalay: We should be consistent with the naming of the templates. The template of `package.json` starts with `_`.  Use the same convention in all the files

sbalay (changes requested):
I guess this is enough for a first approach but we still need to run `create-react-app` here

SKOLZ:
we decided to merge this to have something to work with

PR #5, opened 2017-10-13
Create two type of bootstrap (empty - complete)

NaomiFPassarelli:
- Create two type of bootstrap (empty - complete)
-- Remove each package question
-- Add only two type of kickoff

-- Remove all unused components


generators/app/index.js
```diff
@@ -62,77 +62,18 @@ class GeneratorReact extends Generator {
...
       {
-        type: "confirm",
-        name: "recharts",
-        message: "Would you like to enable recharts?"
+        type: "list",
+        name: "typeOfBootstrap",
+        message: "What type of bootstrap do you want?",
+        default: "empty",
+        choices: [
+          "empty (react, redux, react-dom, redux-form, redux-beacon, redux-thunk, seamless-immutable, react-redux, react-router, react-router-dom, react-router-redux, react-addons-perf, apisauce, postcss, history, prop-types, lodash, i18next, reselect,)",
```
SKOLZ: we should remove `react-addons-perf` since it is deprecated in react 16

generators/app/templates/src/app/index.js
```diff
@@ -13,9 +12,9 @@ class App extends Component {
 
   render() {
     return (
-      <StyleRoot>
+      <div>
```
SKOLZ: we can remove this div, just leave the Routes element

generators/app/templates/src/app/screens/Dashboard/index.js
```diff
@@ -1,30 +1,18 @@
...
-        <Sidebar /> */}
-        <div style={styles.content}>
-          <Switch>
-            <Route exact path={Routes.HOME} component={Home} />
-            <Route exact path={Routes.LOGIN} component={Login} />
-            <Route render={() => <Redirect to={Routes.HOME} />} />
-          </Switch>
-        </div>
-      </div>
+    <div>
```
SKOLZ: you can remove this div too

SKOLZ (dismissed):
just those minor things and merge it directly

SKOLZ (approved):

PR #14, opened 2017-12-29
run cra inside the same folder - change eslint rules

NaomiFPassarelli:
- run cra inside the same folder 
- change eslint rules

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -5,7 +5,7 @@
 echo "Getting initial dependencies..."
 
 system_has() {
-  type "${PWD##*/}" > /dev/null 2>&1
+  type "$1" > /dev/null 2>&1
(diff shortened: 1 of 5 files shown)


run.sh
```diff
@@ -5,7 +5,7 @@
 echo "Getting initial dependencies..."
 
 system_has() {
-  type "$1" > /dev/null 2>&1
+  type "${PWD##*/}" > /dev/null 2>&1
```
mvbattan: This should be `$1` since it is the first parameter of `system_has`, not the first one of the script.

mvbattan (changes requested):
Check comment

mvbattan (approved):

PR #23, opened 2018-07-07
Create searchBar and remove the components that used radium

nicolobruno:
## Summary

Approach SearchBar component. It contains this:
- Div container with custom style
- Button submit

-  Form redux-form wrapped children (custom)

Remove the components that used radium, discussed with @damfinkel 

https://github.com/damfinkel/react-base-components/pull/6

Should be called this way:

```
<SearchBar
  getElements={getElements}
  className={styles.userForm}
  initialValues={{ status: userState }}
>
  <div className={styles.search}>
    <Field
      component={Input}
      name="name"
      placeholder={i18next.t('SearchUser:namePlaceholder')}
      className="m-right-1"
      type="text"
    />
     <Field
      component={Input}
      name="brand"
      placeholder={i18next.t('SearchUser:brandPlaceholder')}
      className="m-right-1"
      type="text"
    />
    <Field
      component={Select}
      name="status"
      dataOptions={STATES}
      placeholder={i18next.t('SearchUser:statusPlaceholder')}
      className="m-right-1"
      type="text"
    />
  <div>
</SearchBar


```


FrankIglesias (changes requested):
You must add the `Button`.
I'd be a nice to also include the `Text` component.

generators/app/templates/src/app/components/SearchBar/layout.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { reduxForm } from 'redux-form';
+
+import Button from '../Button';
```
FrankIglesias: This is not defined :cat:

FrankIglesias:
- Add a readme with a brief example of its usage

generators/app/templates/src/app/components/SearchBar/layout.js
nicolobruno: 😅

generators/app/templates/src/app/components/SearchBar/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import SearchBar from './layout';
+
+function SearchBarContainer({ textButtonSearch, className, initialValues, getElements }) {
+  const handleSubmit = formValues => getElements(formValues);
+  return (
```
damfinkel: /n before return

generators/app/templates/src/app/components/SearchBar/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import SearchBar from './layout';
+
+function SearchBarContainer({ textButtonSearch, className, initialValues, getElements }) {
```
damfinkel: Make this a class inheriting from Component, and make handleSubmit an instance function. This way the handleSubmit method isn't created in each render. No need to separate layout, it's simple enough to leave everything in the index

generators/app/templates/src/app/components/SearchBar/layout.js
```diff
@@ -0,0 +1,30 @@
...
+  textButtonSearch: PropTypes.string,
+  className: PropTypes.string
+};
+
+SearchBar.defaultProps = {
+  className: '',
+  textButtonSearch: 'Buscar'
+};
+
+export default reduxForm({
```
damfinkel: I wouldn't couple this with reduxForm. Let's leave it so the user may not use reduxForm, and eventually if they want to, they can do `reduxForm({ form: MY_FORM })(SearchBarContainer)`.

generators/app/templates/src/app/components/SearchBar/constants.js
```diff
@@ -0,0 +1 @@
+export const SEARCH_FORM = 'SEARCH_FORM';
```
damfinkel: remove this, so we allow the user to define any number of form names instead of only this one, and for the same reasons as I stated in the other comment about reduxForm

damfinkel (changes requested):
1) We also need to remove the copying of the deleted files and add the new files in the copyTemplateFiles step of the installation
2) Try the bootstrap before sending the next PR, so we're sure this doesn't break
3) Can we upload a screenshot of how it looks in the screen that the bootstrap renders by default?

nicolobruno:
Screenshot the component generated local.

captura de pantalla 2018-07-20 a la s 20 48 35

Updated README

damfinkel (approved):

damfinkel:
@FrankIglesias This PR is for adding the SearchBar component. Button, Text and other components will be added in future PRs

generators/app/templates/src/app/components/SearchBar/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function SearchBarContainer({ textButtonSearch, className, buttonClassName, handleSubmit, children }) {
+  return (
+    <div className={className}>
+      <form onSubmit={handleSubmit}>
```
FrankIglesias: attach  the classname to the form instead

FrankIglesias (changes requested):

FrankIglesias (approved):

PR #26, opened 2018-08-02
Update lint-diff to ignore removed files

holywyvern:
## Summary

- changed lint-diff script to take into account removed files

## Notes

This change will help to avoid the problem when lint-diff tries to check for a deleted file with eslint and fail.


damfinkel (approved):
nice!

PR #27, opened 2018-08-03
added storybook as dependency and simple documentation for SearchBar & TextArea

damfinkel:
## Summary

Added https://github.com/storybooks/storybook, which enabled us to have documentation on components as we create them. 
Also added simple examples for TextArea and SearchBar.

## Screencast

svl-react

### DIFF
--- a/generators/app/tasks/installDependencies.js
+++ b/generators/app/tasks/installDependencies.js
@@ -31,7 +31,8 @@ const DEV_DEPENDENCIES = [
   'prettier',
   'prettier-eslint',
   'react-scripts',
-  'wolox-react-scripts'
+  'wolox-react-scripts',
+  '@storybook/react'
 ];
 
 /**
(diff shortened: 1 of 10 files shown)


generators/app/tasks/installDependencies.js
```diff
@@ -31,7 +31,8 @@ const DEV_DEPENDENCIES = [
   'prettier',
   'prettier-eslint',
   'react-scripts',
-  'wolox-react-scripts'
+  'wolox-react-scripts',
+  '@storybook/react'
```
mvbattan, line 35: Shouldn't this be listed as optional ? What do you think guys ?

damfinkel: The way I pictured this is that, after you generate a project, you can run `npm run storybook` and already have examples of all default components the bootstrap gives you. And eventually, add each component in the storybook as you code.

wfolini (approved):

nicolobruno (approved):

mvbattan (approved):

PR #29, opened 2018-08-14
Added equalizer, margins, layout and components scss

nicolobruno:
## Description

- Added https://github.com/Wolox/equalizer for reset the default styles of the browser.
- Added margins.scss, layout.scss and components.scss for to generalize classes of flex-box, margins and width.

Use it in this way:

```js
<div className="row center middle full-width">
  <Input
    name={fieldName}  
    type="text"
    className={`row center m-top-2 ${styles.input}`}
  />
</div>
```

### DIFF
--- a/generators/app/templates/src/scss/layout.scss
+++ b/generators/app/templates/src/scss/layout.scss
@@ -1,0 +1,137 @@
+.column {
+  display: flex;
+  flex-direction: column;
+
+  &.top {
+    justify-content: flex-start;
+  }
+
+  &.middle {
+    justify-content: center;
+  }
+
+  &.bottom {
+    justify-content: flex-end;
+  }
+
+  &.space-between {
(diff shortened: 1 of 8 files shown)


generators/app/templates/src/scss/layout.scss
```diff
@@ -0,0 +1,137 @@
+.column {
```
FrankIglesias, line 1: It's ok, in the future we should add layout as a dependency

generators/app/templates/src/scss/margins.scss
```diff
@@ -0,0 +1,77 @@
+$scale: 10;
```
FrankIglesias: why not adding also responsive margins?

damfinkel (approved):

FrankIglesias (approved):

PR #31, opened 2018-09-06
Update array utilities

holywyvern:
## Summary

- Added `arrayOfIndices` as an utility function for arrays.

## Notes

This is to avoid the use of for-loops when you have to map a set of indices.


damfinkel (approved):
nice!

nicolobruno (approved):

PR #33, opened 2018-09-19
Implement React Rewire Wolox and other improvements

wfolini:
## Summary
Introducing in this pull request: 

* ~~[react-app-rewire-wolox](https://github.com/Wolox/react-app-rewire-wolox) implementation~~
* ~~[Hot reloading](https://github.com/gaearon/react-hot-loader) implementation with [react-app-rewire-hot-loader](https://github.com/cdharris/react-app-rewire-hot-loader)~~
* Allow to generate templates files from .ejs template files
* Replaced deprecated [react-router-redux](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux) with [connected-react-router](https://github.com/supasate/connected-react-router)
* Implemented dummy Home screen
* Force override conflicting files
* Delete create-react-app unnecessary files
* Added register worker
* Other improvements, refactors and fixes

### [OUTDATED] Hot reloading sample.
sep -19-2018 10-39-26


generators/app/index.js
```diff
@@ -8,6 +8,19 @@ const PROMPTS = require('./prompts');
 const { KICKOFF_MESSAGE } = require('./constants');
 
 class GeneratorReact extends Generator {
+  constructor(args, opts) {
+    super(args, opts);
```
mvbattan: I'd write
```
constructor(...args) {
  super(...args);
  ...
}
```
To generalize this usage if needed

generators/app/tasks/installDependencies.js
```diff
@@ -33,7 +33,8 @@ const DEV_DEPENDENCIES = [
   'prettier',
   'prettier-eslint',
   'react-scripts',
-  'wolox-react-scripts',
+  'react-hot-loader',
```
mvbattan, line 46: 👏

generators/app/tasks/utils.js
```diff
@@ -1,7 +1,38 @@
 module.exports.copy = function copy(src, des) {
+  if (!this.fs) {
+    throw new Error('File utils functions needs to be binded to the generator context');
+  }
```
mvbattan, line 4: I'd extract this validation into an aux function that returns a function. What do you think?

generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -10,6 +10,7 @@ import * as Routes from './constants';
 import styles from './styles';
 
 function AppRoutes() {
+  <%# TODO: Handle ConnectedRouter connection according redux feature %>
   return (
     <ConnectedRouter history={history}>
```
mvbattan, line 14: I think that this shouldn't yield 🤔 Github problems ?

generators/app/tasks/configPackageJson.js
```diff
@@ -1,5 +1,4 @@
-const generateRARScript = (command, options) =>
-  `react-app-rewired ${command} --scripts-version wolox-react-scripts${options ? ` ${options}` : ''}`;
+const generateRARScript = (command, options) => `react-app-rewired ${command} ${options ? `${options}` : ''}`;
```
damfinkel: `react-app-rewired ${command} ${options || ''}`

generators/app/constants.js
```diff
@@ -78,15 +71,17 @@ module.exports.TEMPLATE_FILES = [
...
   `${UTILS_PATH}/array.js`,
+  `${SCREENS_PATH}/Dashboard/screens/Home/assets`,
+  `${SCREENS_PATH}/Dashboard/screens/Home/assets/logo.svg`,
   `${SCREENS_PATH}/Dashboard/screens/Home/index.js`,
+  `${SCREENS_PATH}/Dashboard/screens/Home/styles.scss`,
   `${SCREENS_PATH}/Dashboard/index.js`,
   `${COMPONENTS_PATH}/Routes/components/AuthenticatedRoute.js`,
   `${COMPONENTS_PATH}/Routes/constants.js`,
-  `${COMPONENTS_PATH}/Routes/index.js`,
+  `${COMPONENTS_PATH}/Routes/styles.js`,
```
FrankIglesias: Why leaving the styles file?

generators/app/templates/src/scss/margins.scss
```diff
@@ -1,5 +1,6 @@
 $breakpoint-tablet: 768px;
 $breakpoint-desktop: 1024px;
+$scale: 1;
```
damfinkel: isn't this too low? Maybe 2/4?

damfinkel (commented):
Awesome!!!

generators/app/templates/src/redux/store.js
```diff
@@ -1,25 +1,28 @@
-import { createStore, applyMiddleware, compose, combineReducers } from 'redux';
-import thunk from 'redux-thunk';
-import { routerReducer as router, routerMiddleware } from 'react-router-redux';
-import createHistory from 'history/createBrowserHistory';
+import { connectRouter, routerMiddleware } from 'connected-react-router';
```
FrankIglesias, line 1: We might need to add the fetchMiddleware from redux-recompose I think

generators/app/templates/src/index.js
```diff
@@ -1,7 +1,29 @@
...
+  ReactDOM.render(
+    <AppContainer>
+      <App />
+    </AppContainer>,
+    document.getElementById('root')
+  );
+};
+
+// Render once
+render(App);
```
mvbattan, line 20: We aren't using this param

generators/app/templates/src/index.js
```diff
@@ -1,7 +1,29 @@
...
+
+registerServiceWorker();
+
+// Webpack Hot Module Replacement API
+if (module.hot) {
+  module.hot.accept('./app', () => {
+    render(App);
+  });
```
mvbattan, line 28: I'd extract this configuration into another file. What do you think ?

generators/app/templates/src/index.js
mvbattan: Why we need to render this once ? Does this relate with WHMR API ?

generators/app/templates/src/redux/store.js
```diff
@@ -1,25 +1,28 @@
...
 
 const middlewares = [routerMiddleware(history)];
 const enhancers = [];
 
 /* ------------- Thunk Middleware ------------- */
 middlewares.push(thunk);
 
-// TODO Add this if you need it
+/* ------------- Connected React Router Middleware ------------- */
+middlewares.push(routerMiddleware(history));
```
mvbattan: We are pushing this middleware but it already exists in `middlewares`.
I propose to rewrite this as:
```
const middlewares = [thunk, routerMiddleware(history)];
```
So we avoid pushing things to that list. What do you think ?

generators/app/templates/src/redux/store.js
mvbattan: If we are going to use r-r v2.0, we'll also need `configureMergeState` and optionally, `wrapCombineReducers`.

generators/app/templates/src/scss/margins.scss
wfolini: Honestly I do not know, I just put this variable because without, the app 💥

generators/app/constants.js
wfolini: This file was already in the bootstrap, do not modify it. I think that this file shouldn't exist but it doesn't enter into the scope of this pr

generators/app/templates/src/redux/store.js
wfolini: Yeah! There is a lot of work about redux in the bootstrap. Since will be a optional feature, a lot of components and integrations should be refactored. But again, doesn't enter into the scope of this pr

generators/app/templates/src/app/components/Routes/index.ejs
wfolini: ¯\_(ツ)_/¯

generators/app/templates/src/index.js
wfolini: The documentation indicates that it should be implemented in this way
https://github.com/gaearon/react-hot-loader#api

generators/app/templates/src/index.js
wfolini: 🤔 How?

generators/app/templates/src/redux/store.js
wfolini: I didn't realize. good idea!

generators/app/templates/src/scss/margins.scss
damfinkel: Talking with @SKOLZ we'd put 4, at least for now. We can later see how it plays out and change it

wfolini: Great!

generators/app/constants.js
FrankIglesias: However the diff shows that you deleted the file, so you can remove the styles file also

mvbattan: @FrankIglesias `Routes/index.js` appears below 😇

wfolini: Exactly! Check `TEMPLATE_FILES` const below

generators/app/constants.js
```diff
@@ -78,15 +71,17 @@ module.exports.TEMPLATE_FILES = [
   UTILS_PATH,
   `${UTILS_PATH}/colors.js`,
   `${UTILS_PATH}/array.js`,
+  `${SCREENS_PATH}/Dashboard/screens/Home/assets`,
```
pabloccid, line 84: You would be copying a folder here. Does this work? If it does, is it necessary to copy `logo.svg` too?

wfolini: @pabloccid I did not realize this. However, the folder and the assets are copying. 🤔 And if this works, maybe we can reduce this list considerably

generators/app/templates/src/index.js
mvbattan: Then I'd generalize `render` function to receive a `Component`. By now, it doesn't receive any param 😊

generators/app/templates/src/index.js
mvbattan: I thought something like:
```
function configureHotReloading(onConfigure) {
  return module.hot && module.hot.accept('~src/app', onConfigure);
}
```

generators/app/constants.js
mvbattan: Maybe there is an option to recursively copy contents

generators/app/templates/src/index.js
mvbattan: I'd also put here those steps that are needed to configure this module, like `render(App)` once from above.

generators/app/templates/src/app/components/Routes/index.ejs
FrankIglesias: I think it yields 'cause of the file extension

SKOLZ (approved):

FrankIglesias (approved):

damfinkel (approved):

generators/app/constants.js
damfinkel: Of this is working, let's make the refactor elsewhere. I'd like to merge this so we can start using rewired

pabloccid (approved):

mvbattan (approved):

generators/app/tasks/configPackageJson.js
```diff
@@ -1,5 +1,4 @@
-const generateRARScript = (command, options) =>
-  `react-app-rewired ${command} ${options ? ` ${options}` : ''}`;
+const generateRARScript = (command, options) => `react-app-rewired ${command} ${options || ''}`;
```
damfinkel:
```
const generateRARScript = (command, options = '') => `react-app-rewired ${command} ${options}`;
```

generators/app/tasks/copyTemplateFiles.js
```diff
@@ -1,50 +1,42 @@
...
   if (this.features.flow) {
     bindedCopy(FLOWCONFIG_PATH.src, FLOWCONFIG_PATH.destination);
   }
 
-  if (this.features.redux) {
-    REDUX_COMPONENTS.forEach(path => bindedCopy(path, path, null, { projectName: this.projectName }));
-  }
-
-  bindedCopy(LINTER_PATH.src, LINTER_PATH.destination);
+  // Try to delete this
```
damfinkel: should we leave this comment?

generators/app/tasks/configPackageJson.js
wfolini: If there isn't options, prints `react-app-rewired comandGoesHere undefined`

generators/app/tasks/copyTemplateFiles.js
wfolini: I will change a little bit this file in next pr :D

damfinkel (approved):

PR #35, opened 2018-10-05
Apply new linter

damfinkel:
## Summary

- Importing new linter rules from [eslint-config-wolox-react](https://github.com/Wolox/eslint-config-wolox-react) which in turn imports general rules from [eslint-config-wolox](https://github.com/Wolox/eslint-config-wolox)
- Set Create React App version to ^1.0.0 until we update dependencies to support CRA v2. The version that is being installed is 1.5.2
- Fixed all linter errors in boostrap

## Known Issues

Becase we are copying some files that use optional dependencies, when the project is generated it can have linter errors. We should only import those files if the dependencies are chosen, or see how we can fix everything to avoid this. For the CRA created files, we should either remove them or run the linter's autofix on them after generating the project.
Here's the card for this: https://trello.com/c/U5jiQWN2/51-que-al-generar-el-proyecto-no-hayan-errores-de-linter


generators/app/tasks/createReactApp.js
```diff
@@ -2,7 +2,7 @@ const runCommand = require('./runCommand');
 
 module.exports.installCRA = function installCRA() {
   return runCommand({
-    command: ['npm', ['install', '--global', 'create-react-app']],
+    command: ['npm', ['install', '--global', 'create-react-app@^1.0.0']],
```
wfolini: I'd use @1.5.2 version, the last after version 2

generators/app/templates/src/app/components/SearchBar/index.js
```diff
@@ -22,17 +22,17 @@ function SearchBarContainer({
...
 
 SearchBarContainer.propTypes = {
-  textButtonSearch: PropTypes.string,
-  className: PropTypes.string,
-  formClassName: PropTypes.string,
+  handleSubmit: PropTypes.func.isRequired,
   buttonClassName: PropTypes.string,
   children: PropTypes.node,
-  handleSubmit: PropTypes.func.isRequired
+  className: PropTypes.string,
```
wfolini, line 28: I think `className` propTypes isn't necessary

generators/app/templates/src/app/components/SearchBar/index.js
```diff
@@ -22,17 +22,17 @@ function SearchBarContainer({
...
+  formClassName: PropTypes.string,
+  textButtonSearch: PropTypes.string
 };
 
 SearchBarContainer.defaultProps = {
-  className: '',
   buttonClassName: '',
+  className: '',
```
wfolini, line 35: Alto this `className` defaultProps`

wfolini (commented):
Great!

generators/app/templates/src/app/components/Routes/components/AuthenticatedRoute.js
```diff
@@ -10,28 +10,35 @@ const DEFAULT_PUBLIC_ROUTE = Routes.LOGIN;
...
-  // device,
+  /*
+   * TODO Add this if you need it
+   * device,
+   */
   isPublicRoute,
   isPrivateRoute,
   initialized,
-  currentUser,
+  // CurrentUser,
```
mvbattan, line 20: Mmm 🤔

generators/app/templates/src/app/components/Routes/styles.scss
```diff
@@ -0,0 +1,3 @@
+.container {
```
mvbattan, line 1: I'd remove `styles.js` if this one will replace it

generators/app/templates/src/app/components/TextArea/styles.scss
```diff
@@ -0,0 +1,5 @@
+.textarea {
```
mvbattan, line 1: Same with this `styles.js`

generators/app/templates/src/config/numeral.js
```diff
@@ -1,13 +1,13 @@
...
+
 numeral.register('locale', 'es-ar', {
   delimiters: {
     thousands: '.',
     decimal: ','
   },
-  ordinal: function ordinal(number) {
-    return number === 1 ? 'ro' : number === 2 ? 'do' : number === 3 ? 'ro' : 'to';
-  },
+  ordinal: number => (ORDINALS.length - 1 === number ? ORDINALS[number] : 'to'),
```
mvbattan: `'to'` -> `ORDINALS[3]` ?

generators/app/templates/src/redux/store.js
```diff
@@ -21,7 +23,7 @@ middlewares.push(thunk);
 
 // TODO Add this if you need it
 /* ------------- Analytics Middleware ------------- */
-// middlewares.push(AnalyticsMiddleware);
+// Middlewares.push(AnalyticsMiddleware);
```
mvbattan, line 26: mmm 🤔 Does eslint yield with `// _middlewares.push`?

generators/app/templates/src/utils/array.js
```diff
@@ -4,10 +4,14 @@ import Immutable from 'seamless-immutable';
  * Receives an array of strings, and returns an obj with that strings as properties with that string as value.
  * E.G:
  * stringArrayToObject(['A', 'B', 'C']) // { A: 'A', B: 'B', C: 'C' }
+ * @param {array} actionsArray array of values
+ * @param {string} namespace prefix for the resulting values
+ * @returns {object} (['A', 'B', 'C', 'name']) -> { A: '@name/A', B: '@name/B', C: '@name/C' }
  */
 export function stringArrayToObject(actionsArray, namespace = '') {
```
mvbattan, line 11: `createTypes` from `r-r` is the same function 😜

generators/app/tasks/createReactApp.js
damfinkel: It does, when using `^` it takes the last compatible version, which is 1.5.2 right now. I checked it.

generators/app/templates/src/app/components/SearchBar/index.js
damfinkel: Why? I'm using it in this components's parent div

generators/app/templates/src/app/components/SearchBar/index.js
damfinkel: check the comment above

generators/app/templates/src/app/components/Routes/components/AuthenticatedRoute.js
damfinkel: There's a linter rule that forces all comments to start with capital case. So if I comment this, it capitalizes the user. I commented it because it isn't being used.

generators/app/templates/src/redux/store.js
damfinkel: same issue as above... should I remove that rule that sets every comment as capitalized?

generators/app/templates/src/utils/array.js
damfinkel: r-r is an optional dependency... we could make it so it imports this or r-r dependingof which options the user chooses, but it's beyond this PR's scope. I'm only settings the linter and fixing it's issues here.

generators/app/tasks/createReactApp.js
damfinkel: I'll put ^1.5.2

generators/app/templates/src/app/components/Routes/styles.scss
damfinkel: I did, I removed styles.js

wfolini (approved):

generators/app/templates/src/redux/store.js
mvbattan: IMHO will be a bit more comfortable 😊 What do you think ?

generators/app/templates/src/utils/array.js
mvbattan: I know, I know 😅

wfolini (approved):

PR #37, opened 2018-10-09
added Jenkins CI configuration

damfinkel:
## Summary

Added a jenkins configuration for all boostrapped projects. For the time being, it only runs the linter.

Also, removed 2 filepaths, both `styles.js` which didn't exist but he bootstrap was trying to copy them.

Finally, specified the CRA version to one compatible with 1.5.2, as it was failing to create the app with CRA 2.


wfolini (approved):
👌

wfolini (approved):
👌

PR #38, opened 2018-10-18
Component/radiogroup

carlosbollero:
## Description

A radio input implementation.

 **Props**

| Property | Type | Required? | Default | Description |
|:---|:---:|:---:|:---:|:---|
| className | String | | `''` | Styles to apply to the radio inputs container |
| selectedValue | String | | | Specifies the initial selected value on mounting |
| name | String | | | If associated to a form, the name is submitted with the control's value as part of the form data. It is passed to every `RadioOption` child |
| disabled | Boolean | | `false` | Controls if all the options are disabled |
| onChange | Func | | | Function to call on any option selection change |

`RadioGroup.option` uses `RadioOption` component to render each radio input

**RadioOption props**

| Property | Type | Required? | Default | Description |
|:---|:---:|:---:|:---:|:---|
| className | String | | `''` | Styles to apply to the input and label container |
| inputClassName | String | | `''` | Styles to apply to the input |
| labelClassName | String | | `''` | Styles to apply to the label |
| label | String | | `''` | Input associated label |
| value | String | ✓ | | The respective value to be submitted |
| disabled | Boolean | | `false` | Controls if the option available to select |

## Usage

 ```jsx
<RadioGroup name="movies" onChange={() => console.log('clicked')}>
  <RadioGroup.option label="Antman" value="antman" />
  <RadioGroup.option label="Watchmen" value="watchmen" />
  <RadioGroup.option label="Deadpool" value="deadpool" />
</RadioGroup>
```

## Implementation approach
The usual and simpler implementation would be something like this:
```jsx
<RadioOptions
    options={[
      { value: '1', label: 'op 1' },
      { value: '2', label: 'op 2' },
      { value: '3', label: 'op 3', disabled: true }
    ]}
/>
```
Doing it so, i would have to iterate through every option rendering a radio input for each.
This made me think about some future complications like, what if wanted to render another kind of element (e.g. `<hr>` or `<span>`) in between the inputs? or if i wanted to pass more props to the input i'd have to define them inside an object `{ label, value, name, checked, disabled, inputClassName, labelClassName, ...etc }` which isn't quite a "react-like" approach.

So i decided to use the **"compound components"** pattern, defining `RadioGroup` and `RadioOption` component and connecting them using the ***context*** api and a static field `RadioGroup.option`.
This pattern provides the necessary flexibility to render however you want what's inside the 'RadioGroup' component without thinking if the input would receive needed prop. Also, input props are passed just like any other react component, letting you extend the `RadioOption` props api if required.
```jsx
<RadioGroup onChange={() => console.log('click')}>
   <h3>Options</h3>
   <RadioGroup.option label="The Walking Dead" value="thewalkingdead" />
   <div className={styles.someCoolStyle}>
      <RadioGroup.option label="From Hell" value="fromhell" disabled />
      <RadioGroup.option label="V for a Vendetta" value="vvendetta" />
   </div>
   <span>some comments</span>
   <RadioGroup.option label="Watchmen" value="watchmen" />
</RadioGroup>
```

## Storybook
radio1
radio2
radio3
radio4


docs/RadioGroup.md
```diff
@@ -0,0 +1,34 @@
+### RadioGroup
+
+ **Props**
+
+| Property | Type | Required? | Default | Description |
```
FrankIglesias, line 5: add true false  values in the required column

docs/RadioGroup.md
```diff
@@ -0,0 +1,34 @@
...
+| selectedValue | String | | | Specifies the initial selected value on mounting |
+| name | String | | | If associated to a form, the name is submitted with the control's value as part of the form data. It is passed to every `RadioOption` child |
+| disabled | Boolean | | `false` | Controls if all the options are disabled |
+| onChange | Func | | | Function to call on any option selection change |
+
+`RadioGroup.option` uses `RadioOption` component to render each radio input
+
+**RadioOption props**
+
+| Property | Type | Required? | Default | Description |
```
FrankIglesias, line 17: same

docs/RadioGroup.md
```diff
@@ -0,0 +1,34 @@
...
+| className | String | | `''` | Styles to apply to the input and label container |
+| inputClassName | String | | `''` | Styles to apply to the input |
+| labelClassName | String | | `''` | Styles to apply to the label |
+| label | String | | `''` | Input associated label |
+| value | String | ✓ | | The respective value to be submitted |
+| disabled | Boolean | | `false` | Controls if the option is available to select |
+
+**Usage**
+
+ ```jsx
```
FrankIglesias, line 28: js

generators/app/templates/stories/components.js
```diff
@@ -103,3 +104,54 @@ storiesOf('Checkbox', module).add('With label', () => <Checkbox label="Acepto t
...
+  <RadioGroup>
+    <RadioGroup.option label="option 1" value="1" />
+    <RadioGroup.option label="option 2" value="2" />
+    <RadioGroup.option label="option 3" value="3" />
+  </RadioGroup>
+));
+
+storiesOf('RadioGroup', module).add('With custom styles', () => (
+  <RadioGroup className={styles.radioGroup}>
+    <RadioGroup.option
```
FrankIglesias, line 118: I think this might not be the best application for this.
Why not something like 
```js
import { RadioGroup, RadioOption } from '../src/app/components/RadioGroup';
...
```

and in the RadioGroup.js
```js 
export class RadioGroup {
...
}

export const RadioOption = RadioGroup.option;
```

FrankIglesias (changes requested):

generators/app/templates/stories/components.js
mvbattan: I'd let this as is, since it's magic :sparkles:
Notice that this structure lets `RadioGroup` to receive a prop and share it with specific children without cloning them. By example, by passing a `onChange` prop to `RadioGroup`, all `RadioGroup.option` childrens are aware of it without passing it explicitly 🌈

generators/app/templates/stories/components.js
```diff
@@ -103,3 +104,54 @@ storiesOf('Checkbox', module).add('With label', () => <Checkbox label="Acepto t
...
+    <hr />
+    <RadioGroup.option label="option 2" value="2" />
+    <hr />
+    <RadioGroup.option label="option 3" value="3" />
+  </RadioGroup>
+));
+
+storiesOf('RadioGroup', module).add('With callback', () => (
+  <RadioGroup onChange={action('clicked')}>
+    <RadioGroup.option label="option 1" value="1" />
```
mvbattan, line 153: @FrankIglesias Here is an example 😁

generators/app/constants.js
```diff
@@ -92,6 +92,8 @@ module.exports.TEMPLATE_FILES = [
   `${COMPONENTS_PATH}/TextArea/index.js`,
   `${COMPONENTS_PATH}/Checkbox/index.js`,
   `${COMPONENTS_PATH}/Checkbox/layout.js`,
+  `${COMPONENTS_PATH}/RadioGroup/index.js`,
+  `${COMPONENTS_PATH}/RadioGroup/components/RadioOption.js`,
```
mvbattan: `components/RadioOption`, since that one should be placed in a `index.js` file

generators/app/templates/src/app/components/RadioGroup/components/RadioOption.js
```diff
@@ -0,0 +1,60 @@
...
+function RadioOption({
+  className,
+  inputClassName,
+  labelClassName,
+  value,
+  label,
+  name,
+  checked,
+  disabled,
+  onChange,
```
mvbattan: I'd use a `...rest` prop so `input` becomes a bit more customizable. This way, `value`, `name`, `checked`, `disabled` and `onChange` don't need to be explicitly declared 🤗

generators/app/templates/src/app/components/RadioGroup/components/RadioOption.js
```diff
@@ -0,0 +1,60 @@
...
+  );
+}
+
+RadioOption.defaultProps = {
+  className: '',
+  inputClassName: '',
+  labelClassName: '',
+  label: '',
+  checked: false,
+  disabled: false
```
mvbattan: I think none of these `defaultProps` are needed, since this component should behave the same way with `undefs` 🤔

generators/app/templates/src/app/components/RadioGroup/index.js
```diff
@@ -0,0 +1,58 @@
...
+class RadioGroup extends Component {
+  static option = props => (
+    <Consumer>
+      {({ selectedValue, name, disabled, onChange }) => (
+        <RadioOption
+          {...props}
+          name={name}
+          onChange={onChange}
+          checked={selectedValue === props.value}
+          disabled={disabled || props.disabled}
```
mvbattan, line 17: 👏

mvbattan (commented):
Great application of Compound pattern !!

generators/app/templates/stories/components.js
mvbattan: More important, this approach doesn't break encapsulation

docs/RadioGroup.md
carlosbollero: Why? Github supports jsx syntax hightlightning.
jsx
```jsx
<RadioGroup>
   <RadioGroup.option label="Antman" value="antman" />
</RadioGroup>
```
js
```js

<RadioGroup>
   <RadioGroup.option label="Antman" value="antman" />
</RadioGroup>
```

docs/RadioGroup.md
carlosbollero: We are using this format for the props table

| Property | Type | Required? | Default | Description |
|:---|:---:|:---:|:---:|:---|
| className | String | | `''` | Some style |
| value | Number | ✓ | | Some value |

docs/RadioGroup.md
carlosbollero: same

generators/app/constants.js
```diff
@@ -92,6 +92,8 @@ module.exports.TEMPLATE_FILES = [
   `${COMPONENTS_PATH}/TextArea/index.js`,
   `${COMPONENTS_PATH}/Checkbox/index.js`,
   `${COMPONENTS_PATH}/Checkbox/layout.js`,
+  `${COMPONENTS_PATH}/RadioGroup/index.js`,
+  `${COMPONENTS_PATH}/RadioGroup/components/RadioOption/index.js`,
```
nicolobruno, line 96: It would add `README.md` and `docs/RadioGroup.md`

nicolobruno (changes requested):

damfinkel (approved):
Awesome!

mvbattan (approved):
Looks good to me. Check nicolobruno comment !

generators/app/constants.js
carlosbollero: True but you are already adding them in your Login pr.
I think i will let this as is to avoid conflicts.
Let me know.

nicolobruno (approved):

wfolini (approved):
Woww! This component is excellent! Excellent summary! Congratulations! 🎉 👏

FrankIglesias (approved):

PR #40, opened 2018-11-06
css/scss modules can now be used as camelCase in JS

damfinkel:
## Summary

Overrode the webpack configuration so we can import the css/scss modules as camelcase instead of snakecase. 

Also fixed some small issues with the linter

## Screenshots
Example of imported styles:
screen shot 2018-11-06 at 10 59 07


wfolini (approved):
👌

PR #41, opened 2018-11-07
babelrc enabled and added optional chaining plugin

damfinkel:
## Summary

CRA 2 configuration doesn't enable the Babel config to be overriden. So I modified the webpack config to be able to use the .babelrc. This way, I could add the [optional-chaining proposal](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining)

Note: I had to download the Babel JavaScript plugin for VSCode for the syntax highlighting to work properly

## Screenshots

svl-react

### DIFF
--- a/generators/app/constants.js
+++ b/generators/app/constants.js
@@ -40,6 +40,11 @@ module.exports.FLOWCONFIG_PATH = {
   destination: '.flowconfig'
 };
 
+module.exports.BABELRC_PATH = {
+  src: 'babelrc',
+  destination: '.babelrc'
+};
+
 const COMPONENTS_PATH = 'src/app/components';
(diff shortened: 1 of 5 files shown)


generators/app/constants.js
```diff
@@ -45,6 +45,11 @@ module.exports.FLOWCONFIG_PATH = {
   destination: '.flowconfig'
 };
 
+module.exports.BABELRC_PATH = {
+  src: 'babelrc',
+  destination: '.babelrc'
+};
```
wfolini, line 46: In a next pr i will change this approach, to avoid create an object like this for new config file

wfolini (approved):
Eggcellent! 🥚

PR #42, opened 2018-11-12
Add lazy loading and suspense

pabloccid:
Added lazy loading in routes component.


generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -1,21 +1,26 @@
...
 import { history } from '../../../redux/store';
-import Home from '../../screens/Dashboard';
+import Loading from '../Spinner/components/loading';
 
 import AuthenticatedRoute from './components/AuthenticatedRoute';
 import * as Routes from './constants';
 import styles from './styles.scss';
+
+const Home = lazy(() => import('../../screens/Dashboard'));
+
```
wfolini: Delete this space because the comment below add a extra space when is bootstrapped

generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -1,21 +1,26 @@
...
 
 import { history } from '../../../redux/store';
-import Home from '../../screens/Dashboard';
+import Loading from '../Spinner/components/loading';
 
 import AuthenticatedRoute from './components/AuthenticatedRoute';
 import * as Routes from './constants';
 import styles from './styles.scss';
+
+const Home = lazy(() => import('../../screens/Dashboard'));
```
wfolini, line 12: Are we going to add a func like this for each screen that we add? I'd think of a different approach for this

wfolini (changes requested):
Excellent contribution! ❤️ 👏 👏

generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -1,21 +1,26 @@
...
+
 <%# TODO: Handle ConnectedRouter connection according redux feature %>
 function AppRoutes() {
   return (
     <ConnectedRouter history={history}>
       <div className={styles.container}>
-        <Switch>
-          <AuthenticatedRoute isPrivateRoute exact path={Routes.HOME} component={Home} />
-        </Switch>
+        <Suspense fallback={<Loading />}>
```
damfinkel: what about the minimum time to start showing the spinner?

generators/app/templates/src/app/components/Routes/index.ejs
damfinkel: This is how is done. This replaces the import, so it won't have more lines thatn before, just instead of doing `import Home  from ...` you have to do `const Home = lazy(() => import(...))`

wfolini: Sure, you are right... I didn't notice that the imports aren't here :D

wfolini: Sure, you are right... I didn't notice that the normal imports aren't here :D

generators/app/templates/src/app/components/Suspense/index.js
```diff
@@ -0,0 +1,26 @@
+import React, { Suspense } from 'react';
+import PropTypes from 'prop-types';
+
+import Loading from '../Spinner/components/loading';
+
+import { DEFAULT_MAX_DURATION } from './constants';
+
+function CustomSuspense({ maxDuration, fallback, children }) {
```
wfolini, line 8: 👏

generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -1,22 +1,22 @@
-import React, { Suspense, lazy } from 'react';
+import React, { lazy } from 'react';
 import { ConnectedRouter } from 'connected-react-router';
 import { Switch } from 'react-router-dom';
 
 import { history } from '../../../redux/store';
-import Loading from '../Spinner/components/loading';
+import Suspense from '../Suspense';
+
```
wfolini, line 7: Double space?

wfolini (commented):
👏

generators/app/templates/src/app/components/Suspense/index.js
```diff
@@ -0,0 +1,26 @@
+import React, { Suspense } from 'react';
+import PropTypes from 'prop-types';
+
+import Loading from '../Spinner/components/loading';
```
mvbattan, line 4: `components/Loading` ?

generators/app/templates/src/app/components/Routes/index.ejs
```diff
@@ -1,21 +1,25 @@
...
+import React, { lazy } from 'react';
 import { ConnectedRouter } from 'connected-react-router';
 import { Switch } from 'react-router-dom';
 
 import { history } from '../../../redux/store';
-import Home from '../../screens/Dashboard';
+import Suspense from '../Suspense';
 
 import AuthenticatedRoute from './components/AuthenticatedRoute';
 import * as Routes from './constants';
```
mvbattan: This will be needed to be changed in another PR, avoiding `import *`

mvbattan (approved):
Great !

generators/app/templates/src/app/components/Suspense/index.js
damfinkel: This component already existed. If we want to change it let's make it in another PR

damfinkel (approved):

wfolini (approved):

PR #43, opened 2018-11-12
babel module resolver with basic aliases

damfinkel:
## Summary

Babel Plugin Module Resolver - The return of the plugin

Now that babelrc is enabled again, this plugin is working again, so the aliases are back. Also, updated linter versions and fixed a linter bug.


generators/app/templates/.eslintrc.ejs
```diff
@@ -3,5 +3,20 @@ module.exports = {
   plugins: ['flowtype']<% } %><% if (features.jest) { %>,
   env: {
     jest: true
-  }<% } %>
+  }<% } %><% if (features['babel-module-resolver']) { %>,
+  settings: {
```
wfolini, line 7: Why also in this file? What changes apply?

generators/app/templates/.eslintrc.ejs
```diff
@@ -3,5 +3,20 @@ module.exports = {
...
+    'import/resolver': { 
+      'babel-module': {  
+        alias: {
+          '~components': './src/app/components',
+          '~screens': './src/app/screens',
+          '~config': './src/config',
+          '~constants': './src/constants',
+          '~redux': './src/redux',
+          '~services': './src/services',
+          '~utils': './src/utils'
```
wfolini, line 17: Dunno, I don't like this character `~` because the eslint. Can we change ours wolox eslint regex config to avoid the empty line between each import?

wfolini (commented):
Excellent! I also add these paths to vscode confg file `jsconfig.json` for help to the IntelliSense to imports, autoimports or some other things . Check here https://code.visualstudio.com/docs/languages/jsconfig#_using-webpack-aliases

Example
```js
// jsconfig.json
{
  "compilerOptions": {
    ...
    "paths": {
      "@app/*": ["./src/app/*"],
      ...
    }
  },
  "include": ["src"],
  "exclude": ["node_modules", ".history"]
  ...
}

```

generators/app/templates/.eslintrc.ejs
damfinkel: I've tried everything. If we use @, it thinks it's a dependency with [scope](https://docs.npmjs.com/misc/scope) and eslint's rule `import/no-extraneous-dependencies` shows an error. We shouldn't remove that rule, as it's useful to find imports of dependencies that aren't installed.

Here's an issue about this: https://github.com/benmosher/eslint-plugin-import/issues/807
And here's the open PR to fix this that doesn't seem to be going anywhere for the time being: https://github.com/benmosher/eslint-plugin-import/pull/914

generators/app/templates/.eslintrc.ejs
damfinkel: I have to add it in the babelrc to create the alias, and in the eslintrc so the rule `import/no-unresolved` doesn't show an error. This seems to be a bug fixed in the latest beta, asi this issue points out: https://github.com/tleunen/eslint-import-resolver-babel-module/issues/95

But I'd rather not use a beta version of `eslint-import-resolver-babel-module` for the bootstrap

wfolini (approved):
👏

PR #51, opened 2018-12-06
Dependency specific files

gon-escalante:
Added some seamless-immutable-free files, if this dependency is not selected, it copys the entire folder of files and overwrites the ones that have seamless-immutable as a dependency.

I implemented a deepFreeze to have the immutablility.


pabloccid (approved):

generators/app/tasks/copyTemplateFiles.js
```diff
@@ -30,6 +31,10 @@ module.exports = function copyTemplateFiles() {
     bindedCopy(FLOWCONFIG_PATH.src, FLOWCONFIG_PATH.destination);
   }
 
+  if (!this.features['seamless-immutable']) {
```
damfinkel, line 35: I don't see an `else` for when it does have seamless immutable. I'm guessing this copies the files that depend on seamless-immutable and then, if seamless-immutable wasn't chosen, it replaces them with the files that use deepFreeze.

If that's the case, could we avoid copying the same files twice?

generators/app/tasks/copyTemplateFiles.js
```diff
@@ -20,7 +21,7 @@ module.exports = function copyTemplateFiles() {
 
   FILES_TO_DELETE.forEach(src => bindedDeleteFiles(src));
 
-  FILES.forEach(path => bindedCopy(path, path, null, { projectName: this.projectName }));
+  FILES.forEach(path => bindedCopy(path, path));
```
damfinkel, line 25: I think there's no problem with this, but are you sure?

damfinkel (commented):
Nice work! I like that deepFreeze!

generators/app/tasks/copyTemplateFiles.js
gon-escalante: yes, `copy` receives only 2 params and requires you to bind it so it can get the `projectName` from `this`

generators/app/tasks/copyTemplateFiles.js
gon-escalante: hmm i cant do that because we are copying entire folders, and just some files change when seamless-immutable is selected, i'd have to specify every file that is being copied if i want to skip some, i think i'ts better this way, what do you think?

damfinkel (approved):

PR #57, opened 2019-01-05
overrode webpack eslint configuration so console runs our eslintrc config

damfinkel:
## Summary

Webpack console wasn't using the same config as the project. It was using the react's default config, which produces some errors that our linter don't. I overrode it so it uses our eslintrc.

Also, I added `emitWarnings` so they don't crash the app, which is annoying when developing.

Finally, I removed a comment at the top of the config-overrides file for an old rewired we were going to fix but finally decidided not to.

### DIFF
--- a/generators/app/templates/config-overrides.js
+++ b/generators/app/templates/config-overrides.js
@@ -1,5 +1,3 @@
-// Const rewireWolox = require('react-app-rewire-wolox');
-
 const path = require('path');
 
 const createLoaderMatcher = loader => rule =>
@@ -10,6 +8,8 @@ const oneOfFileLoaders = config => config.module.rules.find(rule => rule.oneOf).
 const cssLoaderMatcher = createLoaderMatcher('css-loader');
 
 const babelLoaderMatcher = createLoaderMatcher('babel-loader');
+
+const eslintLoaderMatcher = createLoaderMatcher('eslint-loader');
 
 const enableBabelRc = config => {
   const fileLoaders = oneOfFileLoaders(config);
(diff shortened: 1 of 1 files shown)


generators/app/templates/config-overrides.js
```diff
@@ -36,7 +36,25 @@ const addCamelCaseToCSSModules = (config, env) => {
   });
 };
 
+const useEslintConfig = config => {
+  config.module.rules.forEach(rule => {
+    if (rule.use) {
+      const eslintUse = rule.use.find(use => eslintLoaderMatcher(use));
+
+      if (eslintUse) {
```
mvbattan, line 44: `if (eslintUse && eslintUse.options)` ?

generators/app/templates/config-overrides.js
```diff
@@ -36,7 +36,25 @@ const addCamelCaseToCSSModules = (config, env) => {
...
+        eslintUse.options = {
+          ...eslintUse.options,
+          baseConfig: undefined,
+          useEslintrc: true,
```
mvbattan, line 48: Is `eslintUse.options` immutable ? Do we need to copy and modify it ?

generators/app/templates/config-overrides.js
```diff
@@ -1,5 +1,3 @@
-// Const rewireWolox = require('react-app-rewire-wolox');
```
mvbattan, line 1: This change is not being described at summary

generators/app/templates/config-overrides.js
damfinkel: webpack config is mutable, there's no problem in that sense. Webpack receives an object describing the configuration, rewired just captures that object before actually using it and modifies it before running the start script

generators/app/templates/config-overrides.js
damfinkel: This takes the webpack object form the CRA configuration and overrides all options on purpose. If you look at [this line](https://github.com/facebook/create-react-app/blob/2741781bc6095a43bb79cc259a416075efb915ff/packages/react-scripts/config/webpack.config.js#L311 ) in the config, the options object exists.

If react changes the configuration, we should probably change it too. And if for some reason they remove the `options` in eslintUse, doing `...undefined` doesn't break, so it won't break this either.

FrankIglesias (approved):

mvbattan (approved):

pabloccid (approved):

PR #59, opened 2019-01-11
replaced rewired with rescripts

damfinkel:
## Summary

react-app-rewired is now "lightly" maintained since CRA 2.0. As a result, after react-scripts updated to 2.1.1 and merged webpack.dev.config and webpack.prod.config into one file, react-app-rewired stopped working. We set react-scripts to 2.1.0 for the bootstrap to work, but npm claims it has a "high security vulnerability" and the fix is to update.
The react-app-rewired docs has other alternatives, and I chose [rescripts](https://github.com/rescripts/rescripts). The bootstrap works exactly the same with this package, and the changed were minimal

### DIFF
--- a/generators/app/constants.js
+++ b/generators/app/constants.js
@@ -63,10 +63,14 @@ module.exports.FLOWCONFIG_PATH = {
   destination: '.flowconfig'
 };
 
+module.exports.RESCRIPTS_PATH = {
+  src: 'rescriptsrc.js',
+  destination: '.rescriptsrc.js'
+};
+
 module.exports.REDUX_COMPONENTS = [`${COMPONENTS_PATH}/Field/index.js`];
 
 module.exports.FILES = [
-  'config-overrides.js',
   'pull_request_template.md',
   'README.md',
   'src/index.js',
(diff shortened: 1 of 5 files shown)


FrankIglesias (approved):

generators/app/constants.js
```diff
@@ -63,10 +63,14 @@ module.exports.FLOWCONFIG_PATH = {
   destination: '.flowconfig'
 };
 
+module.exports.RESCRIPTS_PATH = {
```
wfolini, line 66: I think that we can found a better approach to avoid to create a new object per new template config file

generators/app/tasks/installDependencies.js
```diff
@@ -46,10 +45,10 @@ const DEV_DEPENDENCIES = [
   'prettier',
   'prettier-eslint',
   'react-hot-loader',
-  'react-app-rewire-wolox',
```
wfolini, line 49: Can u update `react-app-rewire-wolox` o create a similar lib with actual webpack config and rescripts?? 😇

wfolini (approved):

HarryPerez (approved):
Nice! 👍

mvbattan (approved):
`No lo sé Rick`

pabloccid (approved):

PR #71, opened 2019-05-06
added build, deploy and start scripts for each env

damfinkel:
## Summary

Added scripts to validate the presence of the correct `.env`s when building and deploying. Initially, there are 3 possible environments: `development`, `stage` and `master`. The only way to deploy one of them is to be in a branch with the same name as the environment, and to have a `.env` for that environment. The scripts fails when one of these conditions isn't met, explaining the errors.

If any project needs a variation of these (like more envs or countries) the scripts should be changed to handle that. But it's a good start to avoid bad usage (or no usage) of `.env`s

Scripts added:
- start-env (to start any environment passing it as the first parameter, e.g. `npm run start-env stage`
- build
- deploy

## Screenshots

Some examples of the build failing:

Screen Shot 2019-05-06 at 16 46 38
Screen Shot 2019-05-06 at 16 46 50
Screen Shot 2019-05-06 at 16 47 02

### DIFF
--- a/generators/app/templates/jsconfig.json
+++ b/generators/app/templates/jsconfig.json
@@ -4,7 +4,7 @@
     "module": "commonjs",
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
-    "baseUrl": ".",
+    "baseUrl": "src",
     "paths": {
       "~app/*": ["./src/app/*"],
       "~components/*": ["./src/app/components/*"],
--- a/generators/app/templates/scripts/deploy.js
+++ b/generators/app/templates/scripts/deploy.js
@@ -1,0 +1,36 @@
+const { exec } = require('child_process');
+const { spawn } = require('child_process');
+
+const argv = require('minimist')(process.argv.slice(2));
+
+const { error, success, validateArgs } = require('./utils');
+
+const env = argv._[0];
+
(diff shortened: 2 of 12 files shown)


generators/app/templates/scripts/deploy.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  const currentBranch = stdout.trim();
+
+  if (env !== currentBranch) {
+    error(`Environment ${env} does not match current branch ${currentBranch}`);
+    process.exit(1);
+  }
+
+  const build = spawn('npm run build', [env], { stdio: 'inherit', shell: true });
```
HarryPerez, line 25: What if `Eslint` fails? Does the build process fail?

generators/app/templates/jsconfig.json
```diff
@@ -4,7 +4,7 @@
     "module": "commonjs",
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
-    "baseUrl": ".",
+    "baseUrl": "src",
```
HarryPerez, line 7: why did you change this?

HarryPerez (approved):
Nice!!! 
tenor-254556570

carlosbollero (approved):

PR #72, opened 2019-05-29
Fix module resolver dependency

carlosbollero:
## Summary

* Add babelrc config for opcional dependency only if selected
* Change babelrc for babelrs.js to check that selection


damfinkel (approved):
Nice work! Please be sure it works in a new project

HarryPerez (approved):

PR #73, opened 2019-05-30
removed jest as it comes with CRA 3 and fixed an app.scss bug

damfinkel:
## Summary

The following changes were made to make CRA 3 work:

- Removed jest as an optional dependency. It's installed with react-scripts, and `npm start` fails if the version doesn't match with the one in react-scripts
- Removed the version limitation from CRA installation
- Fixed an issue with layout.scss

### DIFF
--- a/generators/app/templates/src/scss/margins.scss
+++ b/generators/app/templates/src/scss/margins.scss
@@ -2,6 +2,9 @@
 $breakpoint-tablet: 768px;
 $breakpoint-desktop: 1024px;
 
+$bt-scale: null;
+$bd-scale: null;
+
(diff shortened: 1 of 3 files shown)


generators/app/templates/src/scss/margins.scss
```diff
@@ -2,6 +2,9 @@
 $breakpoint-tablet: 768px;
 $breakpoint-desktop: 1024px;
 
+$bt-scale: null;
```
HarryPerez, line 5: Care with stylelint.

HarryPerez (approved):

carlosbollero (approved):

PR #82, opened 2019-08-30
Update build task

franguerini:
## Summary
- Update build task to the correct one


rlajous (approved):

damfinkel (approved):

damfinkel (changes requested):
We're not using react-app-rewired any more, as it's been "lightly supported" for a while now. We use rescripts now

PR #83, opened 2019-09-04
Added check and echo

rlajous:
## Summary
- Added check if yoeman exist if not it installs it

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -48,9 +48,9 @@ fi
 yo --version
 
 if [ $? -eq 0 ]; then
-  echo YEOMAN installed
+  echo Yeoman installed
 else
-  echo YEOMAN not installed, installing now
+  echo Yeoman not installed, installing now...
   npm install -g yo
 fi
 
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -45,6 +45,15 @@ else
   npm i -g yo generator-react-bootstrap > /dev/null 2>&1
 fi
 
+yo --version
+
+if [ $? -eq 0 ]; then
```
FrankIglesias, line 50: check indentation

FrankIglesias (changes requested):

run.sh
```diff
@@ -45,6 +45,15 @@ else
...
+  echo YEOMAN installed
+else
+  echo YEOMAN not installed, installing now
```
damfinkel: "Yeoman not installed, installing now..."

Add the ... and uppercase is kind of ugly, use "Yeoman" instead of YEOMAN

damfinkel (approved):

FrankIglesias (approved):

PR #86, opened 2019-09-13
Deleted not used dependency

rlajous:
## Summary

-Fixed typescript errors


damfinkel (approved):
please check that the new package-lock is correct

PR #94, opened 2019-12-20
Add typescript to custom

pabloferro:
## Summary

Adds TypeScript to customized bootstrap

* TypeScript config
* Eslint + TypeScript config
* Typed app, routes and login components


generators/app/templates/.eslintrc.ejs
```diff
@@ -16,6 +20,37 @@ module.exports = {
           '~utils': './src/utils'
         }
       }
-    }
-   } <%} %>
+    }<%} %>
+  }<% if (features && features.customized) { %>,
+  rules: {
+    'react/jsx-filename-extension': [1, { 'extensions': ['.js', '.jsx', '.tsx'] }],
```
damfinkel, line 26: shouldn't we use https://github.com/Wolox/eslint-config-wolox-typescript?

generators/app/steps/CustomizableSteps/constants.js
```diff
@@ -89,5 +95,15 @@ module.exports.DEPENDENCIES = [
   'seamless-immutable@^7.1.4',
   'history@^4.7.2',
   'i18next@^13.0.0',
-  'react-spinkit@^3.0.0'
+  'react-spinkit@^3.0.0',
+  'typescript@^3.7.2',
+  '@types/jest@^24.0.23',
```
damfinkel, line 103: I think all the @types packages should be dev dependencies

generators/app/templates/src/app/components/Routes/components/AuthenticatedRoute.tsx
```diff
@@ -74,11 +80,4 @@ AuthenticatedRoute.defaultProps = {
...
 
-AuthenticatedRoute.propTypes = {
-  ...Route.propTypes, // eslint-disable-line react/forbid-foreign-prop-types
-  currentUser: PropTypes.bool,
-  isPrivateRoute: PropTypes.bool,
-  isPublicRoute: PropTypes.bool
-};
-
-export default withRouter(connect()(AuthenticatedRoute));
+export default connect()(AuthenticatedRoute);
```
damfinkel: I think we can remove the `connect` if it's not being used, I know you didn't do it but since you'll be making changes...

package.json
```diff
@@ -48,13 +48,14 @@
...
-    "eslint-plugin-babel": "5.3.0",
     "eslint-config-wolox": "3.0.2",
     "eslint-config-wolox-react": "2.0.0",
-    "eslint-plugin-import": "2.17.3",
+    "eslint-plugin-babel": "5.3.0",
+    "eslint-plugin-import": "2.19.1",
     "eslint-plugin-jsx-a11y": "6.2.1",
     "eslint-plugin-prettier": "3.1.0",
     "eslint-plugin-react": "7.13.0",
+    "eslint-plugin-react-hooks": "^2.3.0",
```
damfinkel: I don0t think we need this here, we need it in the main project but the bootstrap itself doesn't use react

damfinkel (changes requested):

generators/app/templates/.eslintrc.ejs
damfinkel: It's not ready yet, so for the time being it'll be like this

pabloferro:
@damfinkel  your changes look good!

PR #102, opened 2020-01-29
Add jenkins

tomiir:
## Summary
Added jenkins and config files


Jenkinsfile
```diff
@@ -0,0 +1,8 @@
+@Library('wolox-ci') _
+
+node {
+
+  checkout scm
+
+  woloxCi('.woloxci/config.yml');
+}
```
facucompe: Perdón, es un TOC
```suggestion
}

```

.woloxci/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10.11.0-alpine
```
damfinkel: I'd use 12.14.1-alpine, which is the latest stable release: https://nodejs.org/en/

damfinkel (changes requested):

damfinkel (approved):

pabloferro (approved):

carlosbollero (approved):
A comment for further developments.
What about adding a step to create a new project, build and run it, then check there is no errors?
This way we will be sure any change we do, doesn't break future bootstrapped projects.
What do you think? @tomiir @pabloferro @damfinkel

tomiir:
@pabloferro  @damfinkel  I'm merging this but I think what @carlosbollero  makes a lot of sense. Will add a card to the trello board so we don't forget about this.

PR #104, opened 2020-01-31
Refactor checkbox component

lucasnestrojil:
## Summary
- Refactor `Checkbox` component by customize option. Improved hooks and typescript
- I decided delete smart & dumb pattern because in this case is only repeated code. What do you think?
- *New*: added general file `_colors.scss` and two general classes for error layouts
- Update Docs

## Screenshots
checkbox

## Jira Card
https://trello.com/c/wAmJM78o/227-actualizar-componentes-checkbox


generators/app/templates/src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from 'react';
+
+interface Props {
+  checked?: boolean;
+  className?: string;
+  inputClassName?: string;
+  labelClassName?: string;
+  label?: string | undefined;
```
damfinkel: don't think `undefined` is needed, being label optional. `label?: string` is enough

generators/app/templates/src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from 'react';
+
+interface Props {
+  checked?: boolean;
```
damfinkel: checked, name, disabled, required and onChange shouldn't be optional I think. The boolean ones have default values so it isn't necessary, and without onChange you wouldn't be able to change the value so it makes sense to make it mandatory. (Also I'd remove the if(onChange) in the handler too)

generators/app/templates/src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  inputClassName?: string;
+  labelClassName?: string;
+  label?: string | undefined;
+  name?: string;
+  disabled?: boolean;
+  required?: boolean;
+  onChange?: (event: React.FormEvent<Element>) => void;
+}
+
+function CheckboxContainer({
```
damfinkel: I'd change the name to Checkbox, now that there's no layout component. And I'd add an `error` prop that we show below the checkbox like we usually do with inputs, and style it in some default way (red color for the text, red border for the checkbox) and then each project will only have to change those default values and they'll have a fully functional Checkbox. What do you think?

damfinkel (changes requested):
Awesome job! I totally agree with the decision to leave behind the smart/dumb pattern in most cases. I added some comments to improve this component a little bit

generators/app/templates/src/app/components/Checkbox/index.tsx
pabloferro: If we make `onChange` optional, does it make sense to have the `isChecked` state?
Right now the `checked` value is only being used for the initial state. That's also a problem.

pabloferro: We can either remove the state and make onChange required, or keep it optional and use a useEffect to keep the state synced with the prop.

lucasnestrojil: It's true! `name`, `onChange` and `checked` must be mandatory, and local state makes no sense. 
But, `disabled` and `required` should have *false* as default so that the component is less strict. What do you think?

generators/app/templates/src/app/components/Checkbox/index.tsx
lucasnestrojil: I like it

generators/app/templates/src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+interface Props {
+  checked: boolean;
+  className?: string;
+  disabled?: boolean;
+  error?: string;
+  errorClassName?: string;
+  inputClassName?: string;
+  label?: string;
+  labelClassName?: string;
```
carlosbollero, line 11: This props are optional but have a default value.
If they are actually optional, i'd remove the default value, otherwise remove the optional marker.

generators/app/templates/src/app/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  inputClassName = '',
+  label = '',
+  labelClassName = '',
+  name,
+  onChange,
+  required = false
+}: Props) {
+  return (
+    <div className={`column start ${className}`}>
+      <div>
```
carlosbollero, line 32: Do we need this `div`?

carlosbollero (changes requested):

generators/app/templates/src/app/components/Checkbox/index.tsx
lucasnestrojil: I need to put `input` and `label` in the same row and the error message below

damfinkel (approved):

damfinkel:
Please add a gif with the example and update documentation if necessary

generators/app/templates/src/app/components/Checkbox/index.tsx
lucasnestrojil: Yes, this make sense with propTypes. But, in this case I see it well for the prolificity of the component.
If i make mandatory the optional props, the developer have to pass some value obligatorily.
What do u think?

carlosbollero: 👍 it made me think that there was some incongruence defining a prop as optional (so letting anybody omit that prop on purpose) but assigning a default value. I discussed it with @pabloferro and explained me some use cases for that.
So we are ok.

carlosbollero (approved):

PR #105, opened 2020-02-03
Refactor RadioGroup Component

lucasnestrojil:
## Summary
- Refactor `RadioGroup` component with Hooks and TypeScript
- Deleted `createContext()` because i think that is not applicable in this case. There is only one level in the component tree (father and childs). 

## Jira Card
https://trello.com/c/OHngsoCP/225-actualizar-componentes-radiogroup


generators/app/templates/src/app/components/RadioGroup/components/RadioOption/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import { IRadioOption } from './interfaces';
```
pabloferro, line 3: We are naming the interface Props. We should keep a convention for that.

I'd move the interface here and name it `Props`. What do you think?
@damfinkel @lucasnestrojil

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  children,
+  className,
+  inputClassName,
+  labelClassName,
+  label,
+  key,
+  ...rest
+}: IRadioOption) {
+  return (
+    <div key={key} className={className}>
```
pabloferro, line 15: You  shouldn't pass `key` here.

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/interfaces.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export interface IRadioOption {
+  checked?: boolean;
+  children: React.FunctionComponent;
```
pabloferro, line 5:
```suggestion
  children: React.ReactNode;
```

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/interfaces.tsx
```diff
@@ -0,0 +1,16 @@
...
+  className?: string;
+  disabled?: boolean;
+  key: string | number;
```
pabloferro, line 8: Not necessary

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/interfaces.tsx
```diff
@@ -0,0 +1,16 @@
...
+  inputClassName?: string;
+  labelClassName?: string;
+  label?: string | undefined;
```
pabloferro, line 11:
```suggestion
  label?: string;
```

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/interfaces.tsx
```diff
@@ -0,0 +1,16 @@
...
+  name?: string;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  value: string;
+  [rest: string]: any;
```
pabloferro, line 15: Do you need this attribute?

generators/app/templates/src/app/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { IRadioOption } from './components/RadioOption/interfaces';
+
+interface Props {
+  disabled?: boolean;
+  className?: string;
+  selectedValue?: string;
+  name?: string;
+  options?: any[];
+  onChange?: (event: React.FormEvent<Element>) => void;
+  [rest: string]: any;
```
pabloferro, line 13: Do you need this attribute?

generators/app/templates/src/app/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  options = [],
+  onChange
+}: Props) {
+  const [selectedValue, setSelectedValue] = useState<string>(selectedValueProp);
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSelectedValue(event.target.value);
+
+    if (onChange) {
+      onChange(event);
```
pabloferro, line 30: Same comment here about the `onChange` and the `selectedValue`.

`selectedValue` isn't being updated by selectedValueProp.

generators/app/templates/src/app/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+  };
+
+  return (
+    <div className={className}>
+      {options.map((childProps: IRadioOption) => (
```
pabloferro, line 36: If options should be of type `IRadioOption` why don't you define it as an `IRadioOption[]` in the `Props` interface?

pabloferro (changes requested):

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/index.tsx
lucasnestrojil: I have doubts with this. If i need use this interface out of the component. How i have to do?
If i move into `index` i've to change export type. 

`IRadioOption` is used in the render
> {options.map((childProps: IRadioOption[]) => ( ...)

@pabloferro @damfinkel

generators/app/templates/src/app/components/RadioGroup/index.js
```diff
@@ -1,60 +0,0 @@
...
-        <RadioOption
-          {...props}
-          name={name}
-          onChange={onChange}
-          checked={selectedValue === props.value}
-          disabled={disabled || props.disabled}
-        />
-      )}
-    </Consumer>
-  );
```
carlosbollero, line 21: We should discuss about deleting this implementation.
The idea behind this approach is to have the freedom to render any component as a child and not just options, and also style each option as needed.
It's probable that in most cases we don't need to, then we can map RadioGroup.option's inside a RadioGroup but i think it is useful to have that freedom. (Also, that is something we should replicate in other components)
Here #38 , in the pull request's description, there is a brief description and an explanation about why i chose that.
Let me know what you think.
@pabloferro @damfinkel @tomiir  you too

carlosbollero (changes requested):

generators/app/templates/src/app/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <RadioOption
+          {...childProps}
+          key={childProps.value}
```
damfinkel, line 39: I wouldn't use the value as key, maybe better to use the name

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/interfaces.tsx
damfinkel: we can remove this, the interface doesn't limit you to add only the stuff you define.

generators/app/templates/src/app/components/RadioGroup/index.tsx
damfinkel: same here, and besides I don't think we want to enable adding more props here

damfinkel (changes requested):
Please add a gif of this working, and update the documentation and the example for this component if the interface changed

generators/app/templates/src/app/components/RadioGroup/components/RadioOption/index.tsx
damfinkel: You shouldn't need to use that outside the component, this interface has the sole purpose of typing the component's parameters. No one outside here should have to import it normally.
This case is not very usual, where you have an `options` object with items with the possible props. I'm not sure you need to define it though, as it should be able to infer them. When you define `options` you should type each property it has, and those types will have to match the Props in the component in order for TS to work.

generators/app/templates/src/app/components/RadioGroup/index.js
damfinkel: You're right. The nicer solucion here I think is to replace this pattern with a render prop:
```
<div className={className}>
  {options.map(option => (
    children(option); // We should type the return value so that it returns something with the "shape" of a radio button so we know the children aren't anything
  )}
</div>
```

Usage:
```
<RadioGroup options={options}>
  {option => <RadioOption key={option.name} {...option} />} // Or whatever you want to pass the props
</RadioGroup>
```
The only strange thing is that this is so generic that it can be used for other stuff that's not a radio group, which is not what we should aim at. I think it's ok to
What do you think? @carlosbollero

PR #107, opened 2020-02-06
Refactor Suspense component

lucasnestrojil:
## Summary
- Refactor Suspense component 
- Deleted maxDuration prop

## Jira Card
https://trello.com/c/yugeKn7T/229-actualizar-componentes-customsuspense


generators/app/templates/src/app/components/Suspense/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { Suspense } from 'react';
+
+import Loading from '../Spinner/components/loading';
+
+interface Props {
+  fallback?: React.ElementType;
+  children: React.ReactNode;
+}
+
+function CustomSuspense({ fallback = Loading, children }: Props) {
```
carlosbollero, line 10: If fallback is optional then the default value isn't needed. If it is, then remove the optional marker.

carlosbollero (changes requested):

generators/app/templates/src/app/components/Suspense/index.tsx
damfinkel: This doesn't work in TS I think. If you define it as not optional, the user must pass something to match the interface. If that's the case, I'd leave it as it is

damfinkel: @pabloferro can you confirm this?

damfinkel (approved):

generators/app/templates/src/app/components/Suspense/index.tsx
pabloferro: Capi is right.
We should provide default values to optional props or handle the undefined case.

carlosbollero (approved):

PR #108, opened 2020-02-06
Refactor input label

lucasnestrojil:
## Summary
- Renamed component
- Implemented typescript on `FormInput`
- Deleted `TextArea` component and moved into FormInput type handler

### Screenshots

image

## Trello Card
https://trello.com/c/HUTfHiPC/230-actualizar-componentes-inputlabel-textarea


generators/app/templates/src/app/components/FormInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  isTextarea?: boolean;
+  inputType: string;
+  label?: string | object | undefined;
+  labelClassName?: string;
+  name: string;
+  onBlur?: (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  onChange: (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  onFocus?: (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  placeholder?: string;
+  readOnly?: boolean;
```
carlosbollero: Same as before, check if this props are optional or not and remove de default value for the optional ones or the marker for the mandatory ones.

carlosbollero (changes requested):

generators/app/templates/src/app/components/FormInput/index.tsx
damfinkel: As I mentioned in another PR, I don't think this is the case with TS, but I'm waiting pablo's confirmation

generators/app/templates/src/app/components/FormInput/index.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+
+interface Props {
+  className?: string;
+  disabled?: boolean;
+  error?: string | undefined;
```
damfinkel: remove undefined

generators/app/templates/src/app/components/FormInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  errorClassName?: string;
+  inputClassName?: string;
+  isTextarea?: boolean;
+  inputType: string;
+  label?: string | object | undefined;
```
damfinkel: remove undefined

generators/app/templates/src/app/components/FormInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const InputComponent = isTextarea ? 'textarea' : 'input';
+  return (
+    <div className={`column start ${className}`}>
+      {label && (
+        <label htmlFor={name} className={`${labelClassName} m-bottom-1`}>
+          {label}
+        </label>
+      )}
+      <InputComponent
+        className={inputClassName}
```
damfinkel, line 47: I'd like to add here something like this:

```
className={`${inputClassName} ${styles.input} ${error && touched && submitCount > 0 ? styles.error : ''}}
```
```
.input {
  border-bottom: 1px solid $alto;

  &:focus {
    border-bottom: 2px solid blue;
  }

  &.error {
    border-bottom: 2px solid red;
  }
}
```

We did this for a project, the most important part is this one where we set style to focus and error, we also did some other stuff which I don't think it's useful to have here, but here's the code in case you want to check it out: https://github.com/Wolox/amgen-disease-management-react/tree/development/src/app/components/FormInput

damfinkel (changes requested):
Please add the gif showing that it works

generators/app/templates/src/app/components/FormInput/index.tsx
briangerez: @damfinkel  @pabloferro, Could you talk about this?

damfinkel: yes, it's like that. If you want a default value in a prop you need to have it as optional in TS, if you don't it'll expect you to pass it and fail.

generators/app/templates/src/app/components/FormInput/index.tsx
briangerez: We can create a card in Trello for this develop. What do you think? :D @damfinkel

damfinkel: Sure, create the card and add the link here please

damfinkel (approved):

carlosbollero (approved):

PR #109, opened 2020-02-06
The three conventions of hoc added in Spinner component

briangerez:
## Summary

I added two conventions of HOC that it is describe in this [doc](https://reactjs.org/docs/higher-order-components.html).

1. Maximizing Composability: Sometimes HOCs accept only a single argument, for example the wrapped component. Usually accept additional arguments, for example, object of configuration. The change that I added is the withSpinner accept a object with all parameters of  a custom Spinner. 
For example: `withSpinner({ color: '00AAE4' })(Button)`

2. Wrap the Display Name for Easy Debugging: I added the Spinner.displayName for view the WrappedComponent in the Debugging. 😎 

### TypeScript

The TYPE_SPINNER constant was deleted because we imported SpinnerProps from react-spinkit directly. In addition, the names of props in the Loading will use the names established by the library. 😄 

This [doc](https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/HOC.md) explain the HOCs with TS.

## Screenshots

Example for convention 2

Captura de Pantalla 2020-02-06 a la(s) 15 50 24

With loading

loading

### Trello Card
https://trello.com/c/Ee9vior5/226-actualizar-componentes-spinner


generators/app/templates/src/app/components/Spinner/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+interface CustomSpinner extends SpinnerProps {
+  classNameContainer?: string;
+}
+
+export function withSpinner<P>(
+  WrappedComponent: React.ComponentType<P>,
+  customSpinner: CustomSpinner = {}
+): React.FC<WithSpinnerProps & P> {
```
carlosbollero: Just to maintain a convention adopted by some well known libraries, i think we should define our hoc's like this
```js
const withSomething = (/* here the params */) => (/* WrappedComponent */) => ...
```
This way we will be using the hoc the same way as, for example
```js
connect(mapStateToProps)(Component)
// or
compose(withSomeFeature, withOtherFeature)(Component)

// spinner
withSpinner({ color: '#FFF' , ... })(Button);
```
What do you think about this?

carlosbollero (changes requested):

generators/app/templates/src/app/components/Spinner/index.tsx
briangerez: I like!

generators/app/templates/src/app/components/Spinner/utils.ts
```diff
@@ -0,0 +1,2 @@
+export const getDisplayName = <P>(WrappedComponent: React.ComponentType<P>) =>
```
carlosbollero: This is more like a general util instead of one concerning strictly to the spinner.
I'd move it to `templates/src/utils/displayName.js`

carlosbollero (changes requested):

damfinkel (commented):
Very nice!! Could you share a screenshot of this component working?

carlosbollero (approved):

PR #112, opened 2020-02-18
Refactored SearchBar

fedeGiussa25:
## Summary

Refactored SearchBar component

## Jira Card
https://trello.com/c/VaEc7slw/187-actualizar-o-borrar-los-componentes-gen%C3%A9ricos


generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+interface Props {
+  buttonClassName?: string;
+  className?: string;
+  formClassName?: string;
+  textButtonSearch?: string;
+  children?: React.ReactNode;
+  handleSubmit?: (event: React.FormEvent<HTMLFormElement>) => void;
```
carlosbollero: In the previous implementation, this prop was required. Are we sure it isn't required anymore?

carlosbollero (changes requested):

generators/app/templates/src/app/components/SearchBar/index.tsx
damfinkel: Yeah I think it should be required, I don't see why a searchbar shouldn't have a way to submit

generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+interface Props {
+  buttonClassName?: string;
+  className?: string;
+  formClassName?: string;
+  textButtonSearch?: string;
+  children?: React.ReactNode;
```
damfinkel: This should be required as well, it would only be a form with a button otherwise

generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  handleSubmit?: (event: React.FormEvent<HTMLFormElement>) => void;
+}
+
+function SearchBarContainer({
+  textButtonSearch = 'Buscar',
```
damfinkel: I'd remove this default value, or at least i18n it.

damfinkel (changes requested):

.eslintrc.js
```diff
@@ -1,14 +1,14 @@
 module.exports = {
-  extends: ['wolox', 'wolox-react', 'wolox-typescript'],
-  parser: '@typescript-eslint/parser',
+  extends: ["wolox", "wolox-react", "wolox-typescript"],
```
pabloferro, line 2: Why are you changing this quotes?

pabloferro (approved):

generators/app/templates/src/app/components/SearchBar/index.test.js
```diff
@@ -0,0 +1,39 @@
...
+    component = mount(
+      <SearchBar submitButtonText="Search" onSubmit={handleSubmit}>
+        <input />
+      </SearchBar>
+    );
+  });
+
+  afterAll(() => {
+    component.unmount();
+  });
```
carlosbollero: I think we should change this to a `beforeEach` and `afterEach` methods. Mounting the component once and not unmounting it after each test makes each test depend on the previous one execution.

generators/app/templates/src/app/components/SearchBar/index.test.js
```diff
@@ -0,0 +1,39 @@
...
+  describe('when pressing the submit button', () => {
+    beforeAll(() => {
+      const button = component.find('[data-testid="search-bar-submit-button"]');
+      button.simulate('submit');
+    });
+
+    it('calls the submit function', () => {
+      expect(handleSubmit).toHaveBeenCalled();
+    });
+  });
```
carlosbollero: May be the second describe can be included inside the first.
```js
  describe('when component is mounted', () => {
    it('matches last snapshot', () => {
      expect(renderer.create(component).toJSON()).toMatchSnapshot();
    });

    describe('when pressing the submit button', () => {
      beforeAll(() => {
        const button = component.find('[data-testid="search-bar-submit-button"]');
        button.simulate('submit');
      });

      it('calls the submit function', () => {
        expect(handleSubmit).toHaveBeenCalled();
      });
    });
  });
```

generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  submitButtonText,
+  className = '',
+  formClassName,
+  buttonClassName = '',
+  onSubmit,
+  children
+}: Props) {
+  return (
+    <div className={className}>
+      <form onSubmit={onSubmit} className={formClassName}>
```
carlosbollero: Are we keeping the outer `div` just to apply some styles? Can't we join the `div` and `form` styles into one `className` and remove the outer `div`?

generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+interface Props {
+  buttonClassName?: string;
+  className?: string;
+  formClassName?: string;
+  submitButtonText: string;
```
carlosbollero: May be this could be optional. As this is a search bar, some times we use image assets instead of a text for the submit button.

generators/app/templates/src/app/components/SearchBar/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onSubmit,
+  children
+}: Props) {
+  return (
+    <div className={className}>
+      <form onSubmit={onSubmit} className={formClassName}>
+        {children}
+        <button className={buttonClassName} type="submit" data-testid="search-bar-submit-button">
+          {submitButtonText}
+        </button>
```
carlosbollero: What if we want an image instead of a text inside the button? Or both. Shouldn't we add a img prop to support that?

carlosbollero (changes requested):

generators/app/templates/src/app/components/SearchBar/index.test.js
damfinkel: Why? It doesn't depend on the first describe

generators/app/templates/src/app/components/SearchBar/index.tsx
damfinkel: Fixed both this and the next content by making this prop `submitButtonContent`, which accepts anything renderable. It might be an image, a text or whatever, but I think this should be mandatory now. I don't think an empty button makes sense.

carlosbollero (approved):

PR #114, opened 2020-03-06
replaced redux for context, changed some files to ts and added reques…

damfinkel:
…t hook

## Summary

- Replaced redux for context and added contextFactory to create contexts and useSelector and useDispatch hooks
- Removed Field component, as it was coupled with redux-form
- Added useAsyncRequest hook to perform requests in a component
- Added TS to some files and examples of usage for context and hooks
- Refactored AuthenticatedRoute, replaced for RouteItem

## Screenshots

pivot


generators/app/steps/CustomizableSteps/constants.js
```diff
@@ -18,40 +17,37 @@ module.exports.NPMRC_PATH = {
...
   DOCS_README_PATH,
-  REDUX_PATH,
   SCREENS_PATH,
   UTILS_PATH,
   'aws.js',
   'tsconfig.json',
   'tsconfig.paths.json',
   'scripts/deploy.js',
   'src/react-app-env.d.ts',
-  `${COMPONENTS_PATH}/Routes/components/AuthenticatedRoute.tsx`,
```
tomiir, line 35: What happened with the authenticated route?

generators/app/steps/CustomizableSteps/copyFiles.js
```diff
@@ -7,7 +7,10 @@ const { FILES, FILES_TO_DELETE, TEMPLATE_FILES, WITHOUT_SEAMLESS_FILES } = requi
 module.exports = function copyAllFiles() {
   mkdirp(this.destinationPath(`${this.projectName}/src/app/assets/`));
 
+  console.log('Deleting default CRA files...');
   deleteFiles.bind(this)(FILES_TO_DELETE);
+
+  console.log('Adding files...');
```
tomiir, line 13: Are these on purpose?

generators/app/templates/src/app/components/Field/index.js
```diff
@@ -1,22 +0,0 @@
-import React from 'react';
```
tomiir, line 1: Why remove this Field component? Can't it be used with another form library? Seems quite generic to me

generators/app/templates/src/app/components/ProviderWrapper/index.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useReducer } from 'react';
+
+interface Props {
+  Context: React.Context<any>;
+  reducer: React.Reducer<any, { type: any }>;
+  initialState: any;
```
tomiir: I'm not particularly fluent in TS but shouldn't this be an `object` type, rather than `any`? Don't know if TS requires you to make the object shape explicit (which clearly we cannot know) so in that case this would be fine.

generators/app/templates/src/app/components/Routes/components/RouteItem.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Route, Redirect, RouteComponentProps, RouteProps } from 'react-router-dom';
+
+interface Props extends RouteProps {
+  component: React.ComponentType<RouteComponentProps<any>> | React.ComponentType<any>;
+  redirectTo?: string;
+}
+
+function RouteItem({ redirectTo, ...config }: Props) {
```
tomiir, line 9: What happens with public/private routes? Are you delegating the reddirection responsability somewhere? 
What I mean is, for example, what if a non-logged user tries to access a private route? We used to handle that kind of cases inside the route component, how do you propose we do that now? 
I think the best way would be for this to be included in the component, or at least create a HOC/wrapper to handle these changes.

generators/app/templates/src/app/components/Routes/constants.ts
```diff
@@ -0,0 +1,32 @@
...
+const MAIN_PRIVATE_PATH = PATHS.home;
+
+export const ROUTES = [
+  {
+    exact: false,
+    path: PATHS.login,
+    component: Login,
+    title: i18next.t('Routes:loginTitle'),
+    description: i18next.t('Routes:loginDescription'),
+    redirectTo: (user: Nullable<User>) => (user ? MAIN_PRIVATE_PATH : undefined)
```
tomiir, line 22: Okay, it seems you want to handle this here. Why?
If you have 500 routes this becomes unmaintainable

tomiir: We could create a common utility function to validate these things, but it would mean calling it over and over again, when including this behaviour in the Route component would need no maintainance. Maybe I'm missing something? 
Let me know what you think

generators/app/templates/src/app/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export type Action = SetUser | ResetUser;
+
+export const actionCreators = {
+  setUser: (user: User) => ({ type: ActionTypes.SET_USER, payload: user }),
+  resetUser: () => ({ type: ActionTypes.RESET_USER })
+};
+
+export const globalReducer = (state: GlobalState, action: Action): GlobalState => {
```
tomiir: I don't really get what this is for. Is it for handling user data? Shouldn't that have a different name?

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
```diff
@@ -1,9 +1,20 @@
...
+
+import withProvider from '~components/ProviderWrapper';
 
 import logo from './assets/logo.svg';
 import styles from './styles.module.scss';
+import { useSelector, Context, useDispatch } from './context';
+import { reducer, INITIAL_STATE, actionCreators } from './reducer';
 
 function Home() {
+  const foo = useSelector(state => state.foo);
```
tomiir, line 12: I'd add a comment to let a new user know why this is here. Otherwise it might seem a little confusing

generators/app/templates/src/app/screens/Login/index.test.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import renderer from 'react-test-renderer';
+
+import Login from './index';
+
+describe('#Login', () => {
```
tomiir, line 9: Could we maybe abstract this to a generic test? Seems like this in particular would work with any component.  It might be nice to create  these generic test cases! (Do we have a card for this? :thinking: )

generators/app/templates/src/app/screens/Login/index.tsx
```diff
@@ -1,29 +1,37 @@
-import React, { Component } from 'react';
+import React, { useCallback } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { useDispatch } from '~app/context';
+import { actionCreators } from '~app/reducer';
```
tomiir: `as userActions` maybe?

generators/app/templates/src/redux/Auth/actions.js
```diff
@@ -1,60 +0,0 @@
-import { push } from 'connected-react-router';
```
tomiir, line 1: all this behaviour was lost right? It would be nice to add cards to recreate all of this in the corresponding components.

generators/app/templates/src/services/AnalyticsService.js
```diff
@@ -1,13 +0,0 @@
-import { createMiddleware as createAnalyticsMiddleware } from 'redux-beacon';
```
tomiir, line 1: same

generators/app/templates/src/services/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+    initialState = null,
+    withPostFetch,
+    shouldExecuteFetch = true,
+    transformResponse = response => response
+  }: AsyncRequestHookParams<P, D, E>,
+  dependencies: []
+) => {
+  const [state, setState] = useState(initialState);
+  const [loading, setLoading] = useState(shouldExecuteFetch || false);
+  const [error, setError] = useState<Error<E> | null>(null);
```
tomiir: why not include the null case type in the interface definition?

generators/app/templates/src/services/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+        request,
+        onPrefetch: () => {
+          setLoading(true);
+          setState(initialState);
+        },
+        onSuccess: data => {
+          if (activeRequest.current === requestId) {
+            const transformedResponse = data ? transformResponse(data) : undefined;
+            setState(transformedResponse);
+            withPostSuccess?.(transformedResponse);
```
tomiir: don't really understand what this does :thinking:

generators/app/templates/src/services/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+          }
+        },
+        onError: errorInfo => {
+          setError(() => errorInfo);
+          withPostFailure?.(errorInfo);
+        },
+        onPostFetch: response => {
+          setLoading(false);
+          if (response.data) {
+            withPostFetch?.(transformResponse(response.data));
```
tomiir: I don't understand how you intend to use the `transformResponse`. I mean, you always mutate data on success, but you re-mutate it here as well. Shouldn't there be 2 different functions here?

tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
tomiir, line 3: why 5?

tomiir (commented):
Looooooots of questions. Nothing really to  change, at least until I clear some doubts. Sorry about the extensive review, but it was an extensive PR :man_shrugging: 
Excelente work though, this seems super solid !

generators/app/steps/CustomizableSteps/constants.js
damfinkel: I changed it to RouteItem, it's more generic. That component has the potential to receive any condition for being shown or not. For instance, if you have many user types with different access levels

generators/app/steps/CustomizableSteps/copyFiles.js
damfinkel: Yes, I used them to debug but then I thought it improved the output for the user

generators/app/templates/src/app/components/Field/index.js
damfinkel: The only function it had was to spread the `input` prop, which Redux Form provides. If we're not using redux-form, it's not very useful

generators/app/templates/src/app/components/ProviderWrapper/index.tsx
damfinkel: You can specify that something is an object, which is more specific than `any`. Although we could support having a state which is not an object I don't see why we would do that. So I agree, I'll change it 👍

generators/app/templates/src/app/components/Routes/components/RouteItem.tsx
damfinkel: redirectTo is a function which receives a user and returns a path to redirect in case that user doesn't meet the condition to show that route. This way we could not only avoid logged users to go to the login or non-logged users to go to the home, we can check access levels or anything we'd like from the user. And with small changes (receiving something other than a user) we could also have access to some routes depending on other factors that aren't the user if we wanted to. This component now is more flexible than the previous AuthenticatedRoute.

generators/app/templates/src/app/components/Routes/constants.ts
damfinkel: you can return `redirectTo: () => {}` if you wanted to, and that would be a full public route. I don't see why that would not be maintainable. Previously we had a gigantic list of routes in the Routes component, it's kind of the same, right? We need to have this information somewhere

generators/app/templates/src/app/reducer.ts
damfinkel: We'll now have more than one store:
1) Each screen will have its store
2) A Global store to save stuff that we'll need to keep from screen to screen, which is normally the user, app theme of something like that. It's not limited to user data, but I think it'll usually be that.

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
damfinkel: Sure, nevertheless this screen will always be deleted, it's the default "Learn React" screen.

generators/app/templates/src/app/screens/Login/index.test.js
damfinkel: Mmm interesting... I'm not sure, normally the tests won't be this simple. I just wanted to leave a snapshot testing example. This should be more complex depending on how the login works. Not sure that can be extracted, if I eventually want to add more tests in each screen.

generators/app/templates/src/app/screens/Login/index.tsx
damfinkel: As I commented before, this reducer isn't limited to users. So I'd prefer it this way.

generators/app/templates/src/redux/Auth/actions.js
damfinkel: Let's do it when we implement the Login component, what do you think? Since it'll only be used there and in the future sign up I think.

generators/app/templates/src/services/AnalyticsService.js
damfinkel: I don't want this service, we weren't even using it. There's a card for GA though: https://trello.com/c/xba9v0SJ/19-google-analytics-seleccionable

generators/app/templates/src/services/utils.ts
damfinkel: I'm not sure what you mean. If you want to allow devs to use and endpoint and avoid handling the error, I don't agree. In every endpoint we should have a possible error and handle it. I think this will force us to do that, which I believe it's an improvement. Error should never be null, endpoints can always fail.
The only reason that it's nullable here is that it starts as null. And I think the definition should not be null (anything could be null in some contexts, so if we make Error nullable we should make anything nullable and it wouldn't make sense)

generators/app/templates/src/services/utils.ts
damfinkel: 1) I removed the request ids, if that's what you meant here
2) withPostSuccess allows the user to do something with the response data at the moment the request finished
3) transformedResponse is a function you can optionally pass to change the shape of the response

generators/app/templates/src/services/utils.ts
damfinkel: I don't understand what you mean. Since both onPostFetch and onSuccess are different callbacks which receive the response, I can't extract the transformed response, so I must do it in these two places separately. withPostFetch is for doing stuff after the request finished, independantly of whether it failed or not

tsconfig.json
damfinkel: I honestly don't know, but it works. This is the tsconfig file for the bootstrap project. It didn't have one and most TS files broke. So I just copied the one from the generated project (which already existed and worked, I think @pabloferro configured it) so the bootstrap doesn't show strange TS errors. We could check this afterwards

generators/app/templates/src/app/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+  switch (action.type) {
```
mcallegari10: Weren't we going to move the switch to a function?

generators/app/templates/src/app/components/Routes/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ROUTES } from './constants';
+import RouteItem from './components/RouteItem';
+import styles from './styles.module.scss';
+
+function AppRoutes() {
+  const user = useSelector(state => state.user);
+
+  return (
+    <Router>
+      <div className={styles.container}>
```
mcallegari10, line 17: This is an obligatory `div`? If not, I think we should delete it

generators/app/templates/src/services/utils.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+import React, { useEffect } from 'react';
+import { mount } from 'enzyme';
+import { PROBLEM_CODE } from 'apisauce';
+
+import { useRequest, useLazyRequest, Error } from '~services/utils';
+import { TestHook, testHook } from '~utils/tests';
+import { Nullable } from '~utils/types';
+
+type True = true;
+type False = false;
```
mcallegari10: Why though? Why not using type `boolean`? 🤔

README.md
```diff
@@ -99,15 +99,8 @@ Additional to clean libraries are added
...
-|[connected-react-router](https://github.com/supasate/connected-react-router)| A Redux binding for React Router v4.
 |[react-router-dom](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-dom)| DOM bindings for React Router.
-|[redux-recompose](https://github.com/Wolox/redux-recompose)| A Redux utility belt for reducers and actions. Inspired by acdlite/recompose.
-|[redux-form](https://github.com/erikras/redux-form)| A Higher Order Component using react-redux to keep form state in a Redux store.
-|[redux-thunk](https://github.com/reduxjs/redux-thunk)| Middlewere for Redux.
 |[react-router](https://github.com/ReactTraining/react-router)| Routing for React.
-|[redux-beacon](https://github.com/rangle/redux-beacon)| Analytics integration for Redux.
 |[seamless-immutable](https://github.com/rtfeldman/seamless-immutable)| Immutable data structures for JavaScript.
 |[history](https://www.npmjs.com/package/history)| Manage session history anywhere JavaScript runs.
 |[i18next](https://www.i18next.com/)| An internationalization-framework written in and for JavaScript.
```
LucasZibell, line 106: Just a suggestion, if we are going to remove Redux and with that redux-form. Shouldn't we add Formik? Maybe we can have it already installed as a customized library

generators/app/steps/CustomizableSteps/constants.js
```diff
@@ -82,16 +79,8 @@ module.exports.OPTIONAL_DEPENDENCIES = {
 
 module.exports.DEPENDENCIES = [
   'apisauce@^1.0.1',
-  'redux@^4.0.1',
-  'react-redux@^6.0.0',
-  'connected-react-router@^6.0.0',
-  'react-router-dom@^4.3.1',
-  'redux-recompose@^2.1.0',
-  'redux-form@^8.0.4',
```
LucasZibell, line 90: Should we add Formik?

generators/app/steps/CustomizableSteps/constants.js
```diff
@@ -18,40 +17,39 @@ module.exports.NPMRC_PATH = {
   destination: '.npmrc'
 };
 
-module.exports.REDUX_COMPONENTS = [`${COMPONENTS_PATH}/Field/index.js`];
-
 module.exports.FILES = [
   CONFIG_PATH,
   CONSTANTS_PATH,
   DOCS_README_PATH,
-  REDUX_PATH,
```
LucasZibell, line 27: Don't we need to change this for a global Context Folder.? Just for global values, like Auth

generators/app/tasks/fileCreators.js
```diff
@@ -60,7 +67,6 @@ module.exports.createJSConfig = function createJSConfig() {
       '~screens/*': ['./src/app/screens/*'],
       '~config/*': ['./src/config/*'],
       '~constants/*': ['./src/constants/*'],
-      '~redux/*': ['./src/redux/*'],
```
LucasZibell, line 63: If we add a context folder we should add it here

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
```diff
@@ -12,11 +24,11 @@ function Home() {
...
-          Learn React
+          Learn {foo}
         </a>
       </header>
     </div>
   );
 }
 
-export default Home;
+export default withProvider({ Context, reducer, initialState: INITIAL_STATE })(Home);
```
LucasZibell, line 34: I noticed we remove documentation from some components. Should we add a little bit of this component? Just explaining what it does and when to use it

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
```diff
@@ -1,9 +1,21 @@
...
+import { reducer, INITIAL_STATE, actionCreators } from './reducer';
 
 function Home() {
+  // Example of how to use these custom hooks
+  const foo = useSelector(state => state.foo);
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(actionCreators.setFoo('React'));
+  }, [dispatch]);
```
LucasZibell, line 17: I know that you are just using this to set an example. But i think it is rather confusing because here you just need a setState. You really don't want to set "React" in all the context. Maybe we can think of a better example and leave this with setState. For example we could dispatch a list of technologies and use them in a children component like a button and show a warning.

generators/app/templates/src/services/utils.ts
```diff
@@ -0,0 +1,131 @@
...
+  useEffect(
+    () => {
+      sendRequest(payload);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    dependencies
+  );
+
+  return [state, loading, error, sendRequest];
+};
```
LucasZibell: I think these 2 should be on a folder call `Hooks` where you can create your custom hooks. I would also add it to the main src and add an access with `~`
```
components
service
hooks
  | useRequest
     | index.ts
     | constants.ts (in case you need it)
     | test.ts
  | useAsyncRequest
....
```

generators/app/templates/src/app/components/ProviderWrapper/index.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useReducer } from 'react';
+
+interface Props<U extends {}> {
+  Context: React.Context<any>;
+  reducer: React.Reducer<U, { type: string }>;
+  initialState: U;
+}
+
+const withProvider = <T extends any, U>({ Context, reducer, initialState }: Props<U>) => (
```
pabloferro: What is this `extends any` doing?

generators/app/templates/src/services/utils.ts
LucasZibell: Im saying this because we should encourage devs to create their custom hooks which not only is a good practice, it solves lots of problems of code repetition. I think with this we normalize the place to where create them and we already give them 2 examples of how to create custom hooks

generators/app/templates/src/app/components/Routes/paths.ts
```diff
@@ -0,0 +1,5 @@
+export default {
```
pabloferro, line 1: I'd move this to the constants file since it's relevant to the routes definition.

generators/app/templates/jsconfig.json
```diff
@@ -11,7 +11,6 @@
       "~screens/*": ["./src/app/screens/*"],
       "~config/*": ["./src/config/*"],
       "~constants/*": ["./src/constants/*"],
-      "~redux/*": ["./src/redux/*"],
```
LucasZibell, line 14: I don't think it is a discussion for now, but should we use `~` or `@` in Viajeros we had some problems using ~ as it assumed all of those roots were imported from node modules. In amex we used @ and we could avoid the `internal/external` package problem using the following eslint configuration.
```
'import/order': [
      'error', {
        'newlines-between': 'always',
        groups: ['builtin', 'external', ['unknown', 'internal'], 'parent', 'sibling', 'index']
      }
    ]
```

With this eslint can differenciate between @apollo and @components. I think it could be a good improvment, what do you think?

generators/app/templates/src/app/components/Routes/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { useSelector } from '~app/context';
+
+import Suspense from '../Suspense';
+
+import { ROUTES } from './constants';
+import RouteItem from './components/RouteItem';
+import styles from './styles.module.scss';
+
+function AppRoutes() {
```
pabloferro: Can you rename the component to match the folder name?

generators/app/templates/src/app/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+  SET_USER = 'SET_USER',
+  RESET_USER = 'RESET_USER'
+}
+
+export interface SetUser {
+  type: ActionTypes.SET_USER;
+  payload: User;
+}
+
+export interface ResetUser {
```
pabloferro: Do you need to export individual action types?

LucasZibell (changes requested):

generators/app/templates/src/app/reducer.ts
pabloferro: It's a good point.
I'm just thinking out loud, but wouldn't it make sense to have different contexts for different global data?
To avoid unnecessary updates, for example updating consumers of the user when the theme changes.

generators/app/templates/src/app/screens/Dashboard/screens/Home/reducer.ts
```diff
@@ -0,0 +1,42 @@
...
+  SET_FOO = 'SET_FOO',
+  RESET_FOO = 'RESET_FOO'
+}
+
+export interface SetFoo {
+  type: ActionTypes.SET_FOO;
+  payload: string;
+}
+
+export interface ResetFoo {
```
pabloferro: Same as before, do you need to export these?

generators/app/templates/src/services/utils.test.tsx
pabloferro: Or just `true` or `false` directly

generators/app/templates/src/services/utils.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+  {
+    foo: string;
+  } | null,
+  boolean,
+  Nullable<Error<FooError>>,
+  (params?: any) => void
+];
+
+const failureResponse = {
+  ok: false as False,
```
pabloferro:
```suggestion
  ok: false as const,
```
does that work?

generators/app/templates/src/services/utils.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+  originalError: { config: {}, isAxiosError: false, toJSON: jest.fn(), name: '', message: '' },
+  data: { errorData: 'error' },
+  status: 400,
+  headers: {},
+  config: {},
+  duration: 100
+};
+
+const successResponse = {
+  ok: true as True,
```
pabloferro:
```suggestion
  ok: true as const,
```

generators/app/templates/src/services/utils.ts
```diff
@@ -0,0 +1,131 @@
+import { useState, useEffect, useCallback } from 'react';
```
pabloferro: This looks great!

Maybe we can name the file requests or something like that instead of utils?

pabloferro (changes requested):
Nice work!

generators/app/templates/src/app/reducer.ts
pabloferro: It's really hard to type 😢

damfinkel: I think I told you that in order to have strong action typing I had to use switch. We encountered issues trying to move to a reducerCreator with an object, either because of TS limitations or because we just couldn't figure it out. After a while I decided to go with this as a first approach so as to finish

generators/app/templates/src/app/components/Routes/index.tsx
damfinkel: I didn't put this here (I just changed it to TS). Nevertheless, I think it depends on how you look at it. It is not mandatory, but it might be good to have a container for the whole app with the height set to 100vh (or whatever size it should be) so we don't have to do that in every screen's container.

generators/app/templates/src/services/utils.test.tsx
damfinkel: yeah, there's kind o a bug I think in TS with this. The problem is that apisauce has the following definition: 
```ts
export interface ApiErrorResponse<T> {
  ok: false
  problem: PROBLEM_CODE
  originalError: AxiosError

  data?: T
  status?: number
  headers?: {}
  config?: AxiosRequestConfig
  duration?: number
}
export interface ApiOkResponse<T> {
  ok: true
  problem: null
  originalError: null

  data?: T
  status?: number
  headers?: {}
  config?: AxiosRequestConfig
  duration?: number
}
```

The issue is that, for instance if I want to have an object with type ApiOkResponse and I set `true`, it shows the following error:

image

So I had to cast it to a type with the fixed value of true. 

Although I just realized I can do this:
`ok: true as true`

and it appears to work. I'll change it.

Here's an issue related to this, although I don't understand if they're working on it or not: https://github.com/Microsoft/TypeScript/issues/24413

generators/app/templates/src/app/components/ProviderWrapper/index.tsx
damfinkel: I'll change it to {}, I was using it to pass as prop type for the wrapped component

README.md
damfinkel: maybe, but I won't do it in this PR. As of right now, nothing is actually using redux-form, so I'm not removing any functionality. We should think if we want to force Formik, set it as optional dependency or what do we actually want

generators/app/steps/CustomizableSteps/constants.js
damfinkel: same answer

generators/app/steps/CustomizableSteps/constants.js
damfinkel: global context is in ~app/context and it's already being added in this same object:

```
module.exports.FILES = [
  ...
  'src/app/reducer.ts',
  'src/app/context.ts',
  ...
]
```

generators/app/tasks/fileCreators.js
damfinkel: I think the I answered this in the previous comment

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
damfinkel: I think all the docs I removed are for components that aren't present anymore. I like the idea of adding documenttion for this component, but I rather do it in another card so we can close this. I added it to the backlog: https://trello.com/c/ficXXNHQ/241-agregar-documentacion-para-withprovider

generators/app/templates/src/app/screens/Dashboard/screens/Home/index.js
damfinkel: I realize it's not the best example, I just wanted to be expeditive. I don't think this is very important though, given that this component will be removed 100% of the times.

generators/app/templates/src/services/utils.ts
damfinkel: I like it, I'll change it

generators/app/templates/jsconfig.json
damfinkel: I think I had issues with jest for some reason when I tried it to use @. Can you add a card in here https://trello.com/b/sZVeiFgw/departamento-de-frontend so we can do this separatedly?

generators/app/templates/src/app/components/Routes/paths.ts
damfinkel: I had to move it elsewhere because it was generating circular dependencies. Routes constant import components and those components sometimes import paths. If those paths are in Routes, that's a circular dependency. Given that components shouldn't import routes, only path, I think it might be actually better this way.

generators/app/templates/src/app/reducer.ts
damfinkel: probably not... at least not right now. I'll remove it

generators/app/templates/src/services/utils.test.tsx
damfinkel: `ok: false as false` works, I just tested it and I'll change it

generators/app/templates/src/services/utils.ts
damfinkel: I'll take the suggestion from @LucasZibell and move them to a hooks folder to encourage creating custom hooks. I think I'll separate them too.

LucasZibell (approved):

generators/app/templates/tsconfig.paths.json
```diff
@@ -2,15 +2,17 @@
...
       "~services/*": ["services/*"],
       "~components/*": ["app/components/*"],
       "~screens/*": ["app/screens/*"],
-      "~config/*": ["app/config/*"],
-      "~schema/*": ["schema/*"]
+      "~config/*": ["config/*"],
+      "~schema/*": ["schema/*"],
+      "~app/*": ["app/*"],
+      "~hooks/*": ["./src/hooks/*"],
+      "~contexts/*": ["./src/contexts/*"]
```
pabloferro:
```
      "~hooks/*": ["app/hooks/*"],
      "~contexts/*": ["app/contexts/*"]
```

I'd move these to app

pabloferro (changes requested):

generators/app/templates/tsconfig.paths.json
```diff
@@ -11,8 +11,8 @@
       "~config/*": ["config/*"],
       "~schema/*": ["schema/*"],
       "~app/*": ["app/*"],
-      "~hooks/*": ["./src/hooks/*"],
-      "~contexts/*": ["./src/contexts/*"]
+      "~hooks/*": ["./app/hooks/*"],
```
pabloferro:
```suggestion
      "~hooks/*": ["app/hooks/*"],
```

generators/app/templates/tsconfig.paths.json
```diff
@@ -11,8 +11,8 @@
...
+      "~contexts/*": ["./app/contexts/*"]
```
pabloferro:
```suggestion
      "~contexts/*": ["/app/contexts/*"]
```

pabloferro (changes requested):

pabloferro (approved):

mcallegari10 (approved):

tomiir (approved):

PR #116, opened 2020-03-19
Login story

tomiir:
## Summary
Added layout for "Forgot Password" and "Registration" screens.
Layout is simple and generic, no emphasis was made on CSS since this will be completely reconfigured in each proyect, and no functionality other than navigation between screens is provided. 
Modified some tags to use proper headers. Maybe I'm mistaken, let me know.
As a next step it could be a good idea to allow the registration to allow multiple steps each with a corresponding component, but this also depends on the proyect. 

## Screenshot
Screenshot from 2020-03-27 17-08-34
Screenshot from 2020-03-27 17-08-23
Screenshot from 2020-03-27 17-08-14

## Trello Card
https://trello.com/c/39ICVr3d/233-login-v2


generators/app/templates/src/app/screens/RecoverPassword/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+
+import RecoverPassword from './layout';
+
+class RecoverPasswordContainer extends Component {
```
briangerez: It should be a function component

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+
+import Registration from './layout';
+
+class RegistrationContainer extends Component {
```
briangerez: same

generators/app/templates/src/app/screens/Registration/layout.tsx
```diff
@@ -0,0 +1,94 @@
...
+          name={FIELDS.confirmPassword}
+          inputId={FIELDS.confirmPassword}
+          dataFor={FIELDS.confirmPassword}
+          inputType="text"
+          inputClassName={`m-left-1 m-bottom-2 full-width ${styles.input}`}
+          placeholder={i18next.t('Registration:confirmPasswordPlaceholder') as string}
+          handleChange={onInputChange}
+        />
+      </div>
+      <div className={`column center ${styles.sectionContainer}`}>
```
briangerez: This div is identical to RecoverPassword layout. Maybe you could extract it. It is a suggestion, does it look good? 🤔

briangerez (changes requested):

generators/app/templates/src/app/screens/Registration/layout.tsx
tomiir: I thought about this, but since the login and registration almost never look the same it seemed pointless. If anyone else considers that this is necessary I'll gladly do it!

generators/app/templates/src/app/screens/RecoverPassword/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import RecoverPassword from './layout';
+
+function RecoverPasswordContainer() {
+  const handleSubmit = useCallback((event: React.FormEvent<Element>) => {
+    event.preventDefault();
+    // TODO: Implement
+  }, []);
+
+  const handleEmailChange = useCallback((event: React.FormEvent<Element>) => {
+    event.preventDefault();
```
tomiir: added this to avoid ts from alerting errors on unused params !

briangerez:
I think you forgot the screenshots, jaja

generators/app/templates/src/utils/array.ts
```diff
@@ -8,12 +8,15 @@ import Immutable from 'seamless-immutable';
...
  */
-export function stringArrayToObject(actionsArray, namespace = '') {
-  if (actionsArray.some(actionName => !actionName || typeof actionName !== 'string')) {
+export function stringArrayToObject(actionsArray: string[], namespace = ''): { [key: string]: string } {
+  if (actionsArray.some((actionName: string) => !actionName || typeof actionName !== 'string')) {
     throw new Error('Action names must be strings and must not be empty');
   }
 
-  return new Immutable(actionsArray).asObject(actionName => [actionName, `${namespace}:${actionName}`]);
+  return new (Immutable as any)(actionsArray).asObject((actionName: string) => [
```
damfinkel: what's this Immutable as any? Why do you need to do this?

tomiir: Had forgotten about this any but it was because TS was demanding the type for immutable and I couldn't find it. Will look for it though, don't worry

generators/app/templates/src/app/screens/RecoverPassword/index.tsx
carlosbollero: Can't this be configured to be ignored in some config file (may be `tsconfig.json`) instead of adding a line of code?

generators/app/templates/src/app/screens/RecoverPassword/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+    );
+  });
+
+  afterEach(() => {
+    component.unmount();
+  });
+
+  describe('when mounting the application', () => {
+    it('matches last snapshot', () => {
+      expect(renderer.create(component).toJSON()).toMatchSnapshot();
```
carlosbollero, line 26: Made a comment about this in this other pr  https://github.com/Wolox/react-bootstrap/pull/119#discussion_r405060028
Just to discuss, don't need to change it.

generators/app/templates/src/app/screens/RecoverPassword/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+        <RecoverPassword />
+      </RootComponent>
+    );
+  });
+
+  afterEach(() => {
+    component.unmount();
+  });
+
+  describe('when mounting the application', () => {
```
carlosbollero: `"when component is mounted"`

generators/app/templates/src/app/screens/RecoverPassword/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+    // TODO: Implement method
+  }, []);
+  return <RecoverPassword onSubmit={handleSubmit} onEmailChange={handleEmailChange} />;
```
carlosbollero: Didn't we stop using the `index(smart)/layout(dumb)` pattern?

generators/app/templates/src/app/screens/RecoverPassword/layout.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+function RecoverPassword({ onSubmit, onEmailChange }: Props) {
+  return (
+    <form className={`column center full-width m-top-8 ${styles.formContainer}`} onSubmit={onSubmit}>
+      <div className="column center m-bottom-3">
+        <h1 className="m-bottom-1">{i18next.t('RecoverPassword:recoverPassword')}</h1>
+        <h2>{i18next.t('RecoverPassword:recoverPasswordDescription')}</h2>
+      </div>
+      <div className={`column m-bottom-2 ${styles.sectionContainer}`}>
+        <FormInput
```
carlosbollero: Is this outer div needed? It only have one child and the form component is already a flex column. Just to check.

generators/app/templates/src/app/screens/Registration/constants.tsx
```diff
@@ -0,0 +1,3 @@
+import { stringArrayToObject } from '~utils/array';
+
+export const FIELDS = stringArrayToObject(['name', 'surname', 'email', 'password', 'confirmPassword']);
```
carlosbollero: I have my concerns about defining this kind of constants, we usually don't use them in more than one place and we are adding a file that may not be necessary.
Don't need to change this because we are used to doing this but i just wanted to make my point.

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useCallback } from 'react';
+
+import Registration from './layout';
+
+function RegistrationContainer() {
```
carlosbollero, line 12: Same as before about smart/dumb pattern.

carlosbollero (changes requested):

generators/app/templates/src/app/screens/Registration/constants.tsx
tomiir: I see your point. I believe that this is still better than having several literals spread across the code since it allows for easier understanding/maintenance. 
Maybe we could see a way to embed this directly in the component to avoid adding an extra file, but I'm not sure how good that practice is.

carlosbollero (approved):

generators/app/templates/src/app/screens/RecoverPassword/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const handleSubmit = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement
+  }, []);
+
+  const handleEmailChange = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement method
+  }, []);
+
+  return (
+    <form className={`column center full-width m-top-8 ${styles.formContainer}`} onSubmit={handleSubmit}>
```
damfinkel: this top margin should be a padding

generators/app/templates/src/app/screens/RecoverPassword/styles.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  padding: 5px;
+}
+
+.form-container {
+  min-height: 500px;
+}
+
+.section-container {
+  width: 100%;
+  max-width: 280px;
```
damfinkel, line 20: property order

generators/app/templates/src/app/screens/Registration/constants.tsx
damfinkel: I'm fine with this constants, but maybe we could have them in the component to avoid a new file. I think we should be more flexible regarding where to put constants. If the component is small, no need to move this to a constant file unless there are several constants.

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const handleSubmit = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement
+  }, []);
+
+  const handleInputChange = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement method
+  }, []);
+
+  return (
+    <form className={`column center full-width m-top-8 ${styles.formContainer}`} onSubmit={handleSubmit}>
```
damfinkel: same with the margin as before

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <h1 className="m-bottom-1">{i18next.t('Registration:registration')}</h1>
+        <h2>{i18next.t('Registration:registrationExplanation')}</h2>
+      </div>
+      <div className={`row space-between m-bottom-2 full-width ${styles.sectionContainer}`}>
+        <FormInput
+          className="full-width"
+          label={i18next.t('Registration:name')}
+          name={FIELDS.name}
+          inputType="text"
+          inputClassName={`m-right-1 m-bottom-2 full-width ${styles.input}`}
```
damfinkel: this martin right is space between input and its parent. It should be padding in the parent

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+          placeholder={i18next.t('Registration:namePlaceholder') as string}
+          onChange={handleInputChange}
+        />
+        <FormInput
+          className="full-width"
+          label={i18next.t('Registration:surname')}
+          name={FIELDS.surname}
+          inputType="text"
+          inputClassName={`m-left-1 m-bottom-2 full-width ${styles.input}`}
```
damfinkel: prefer using m-right-1 in the labelClassName. We try to avoid margin left and top

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+          onChange={handleInputChange}
+        />
+      </div>
+      <div className={`row space-between m-bottom-2 full-width ${styles.sectionContainer}`}>
+        <FormInput
+          className="full-width"
+          label={i18next.t('Registration:password')}
+          name={FIELDS.password}
+          inputType="password"
+          inputClassName={`m-right-1 m-bottom-2 full-width ${styles.input}`}
```
damfinkel: same about this margin right and the following margin left

generators/app/templates/src/app/screens/Registration/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  padding: 5px;
+  font-size: 14px;
+}
+
+.form-container {
+  min-height: 500px;
+}
+
+.section-container {
+  width: 100%;
```
damfinkel: property order

generators/app/templates/src/app/screens/Registration/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  border-radius: 5px;
+  height: 40px;
+}
+
+.input {
+  height: 30px;
+  border: solid 1px;
+  border-radius: 5px;
+  padding: 5px;
+  font-size: 14px;
```
damfinkel, line 14: property order

generators/app/templates/src/app/screens/Registration/styles.module.scss
```diff
@@ -0,0 +1,22 @@
+.button {
+  border: solid 1px;
+  border-radius: 5px;
+  height: 40px;
+}
+
+.input {
+  height: 30px;
```
damfinkel: property order

damfinkel (changes requested):

generators/app/templates/src/app/screens/Registration/index.tsx
tomiir: It's actually used to separate the `name` input container from the `surname` one. Would you prefer for me to use the `className` prop and add a padding there?

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  const handleSubmit = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement
+  }, []);
+
+  const handleInputChange = useCallback((event: React.FormEvent<Element>) => {
+    // TODO: Implement method
+  }, []);
+
+  const FIELDS = stringArrayToObject(['name', 'surname', 'email', 'password', 'confirmPassword']);
```
damfinkel: this is a constant, move it outside the component so as to not redefine it in each render

generators/app/templates/src/app/screens/Login/layout.tsx
```diff
@@ -16,11 +14,12 @@ interface Props {
 }
 
 function Login({ onEmailChange, onPasswordChange, onLogin }: Props) {
+  const FIELDS = stringArrayToObject(['email', 'password']);
```
damfinkel: same here

generators/app/templates/src/app/screens/Registration/index.tsx
damfinkel: nope, this is ok then, my bad

generators/app/templates/src/app/screens/Registration/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+          className="full-width"
+          label={i18next.t('Registration:name')}
+          name={FIELDS.name}
+          inputType="text"
+          inputClassName={styles.input}
+          placeholder={i18next.t('Registration:namePlaceholder') as string}
+          onChange={handleInputChange}
+        />
+        <FormInput
+          className="full-width m-left-1"
```
damfinkel, line 39: replace margin left with right in the previos element

damfinkel (changes requested):

briangerez (approved):

carlosbollero:
@damfinkel if you are ok with these changes, we shall merge this pr.

mcallegari10 (approved):

PR #117, opened 2020-03-26
Google Analytics

carlosbollero:
* add ga option into generator
* add ga script to index.html template

## Summary

Add Google Analytics script optionally, depending on user choice.
The generator prompts if `ga` should be added or not. If so, asks for the corresponding ID.

*Set the `REACT_APP_GA_TRACK_ID` environment variable in the corresponding .env*

## Screenshots

| | Screen |
|:--:|:--:|
| __Menu__ | image |
| __Result__ | image |
| __.env__ | image |
| __App running__ | image |

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -39,7 +39,7 @@ module.exports.MAIN_PROMPTS = [
     when: values => values.ga,
     type: 'input',
     name: 'gaID',
-    message: 'Enter your Google Analytics ID (leave blank to add it later)',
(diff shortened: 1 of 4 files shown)


pabloferro:
Looks good but I'd get the ID from an env var, what do you think?

carlosbollero:
> Looks good but I'd get the ID from an env var, what do you think?

@pabloferro Agree! I'll add that in a minute.

damfinkel (approved):

PR #119, opened 2020-03-30
RadioGroup changed to use context and TS

damfinkel:
## Summary

Changed RadioGroup so it uses TypeScript. Also, changed the logic a but so it mainly supports RadioOption as children but could eventually support any children. The data passed from RadioGroup to its children is passed through a Context.

Also, hotfixed a problem with TS inference detected when using the reducer/context structure in another project.

## Screenshots

pivot

image


generators/app/templates/src/app/components/RadioGroup/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+    it('all the inputs are disabled', () => {
+      const selectedInput1 = getInput(component, 'option1');
+      const selectedInput2 = getInput(component, 'option2');
+      expect(selectedInput1.props().disabled).toBeTruthy();
+      expect(selectedInput2.props().disabled).toBeTruthy();
+    });
+
+    it('matches last snapshot', () => {
+      expect(renderer.create(component).toJSON()).toMatchSnapshot();
```
carlosbollero, line 37: May be we should consider using [enzyme-to-json](https://github.com/adriantoine/enzyme-to-json) to avoid mixing testing libraries. 
This way,
```js
expect(component).toMatchSnapshot();
```
would work properly.
It adds another dependency but makes the tests a little easier to understand and devs don't need to be thinking which library is useful for each situation.
I tried it and every test worked correctly (snapshots need to be updated).
Tell me what you think @damfinkel @pabloferro 
This is the basic config that needs to be added
```js
// npm install --save-dev enzyme-to-json
"jest": {
  "snapshotSerializers": [
    "enzyme-to-json/serializer"
  ]
}
```

carlosbollero (approved):
The comment i did doesn't block this pr to be merged as it's ok to me.
Check conflicts, you may find a typo i introduced in `fileCreators.js` file.

generators/app/templates/src/app/components/RadioGroup/index.test.js
tomiir: this seems waaay cleaner. How heavy is the dependency? We'd need to refactor most if not all tests too though

damfinkel: I like it, let's add a task to do so

pabloferro (approved):

PR #123, opened 2020-04-03
Form input styles

briangerez:
## Summary

I added input styles for focus and errror.

I removed the generators/app/templates/src/utils/colors.js file because I think thay it is not used.

I updated the readme with the modification of components.

Note: The screenshots of login are in illustrative mode. The login did not really look like this.

## Screenshots

Captura de Pantalla 2020-04-02 a la(s) 18 55 33

Captura de Pantalla 2020-04-02 a la(s) 18 56 12

DOC:
image

### Trello Card
https://trello.com/c/ubkhzBrk/240-agregar-estilos-de-error-para-inputlabel


damfinkel (approved):

README.md
```diff
@@ -56,13 +56,13 @@ Choose your preferred libraries for your project and create a new React project
 
 ###### Components
 - Checkbox
-- InputLabel
+- FormInput (input and textarea)
```
carlosbollero, line 59: Shouldn't we update the corresponding doc?
https://github.com/Wolox/react-bootstrap/blob/form-input-styles/generators/app/templates/docs/InputLabel.md

README.md
```diff
@@ -56,13 +56,13 @@ Choose your preferred libraries for your project and create a new React project
...
+- ProviderWrapper
 - RadioGroup
 - Routes
 - SearchBar
 - Spinner
-- Suspense 
-- TextArea
```
carlosbollero, line 65: And remove this doc as well?

generators/app/templates/src/app/components/FormInput/index.tsx
```diff
@@ -33,9 +37,13 @@ function FormInput({
...
   onFocus,
   placeholder = '',
-  readOnly = false
+  readOnly = false,
+  touched,
+  submitCount
 }: Props) {
   const InputComponent = isTextarea ? 'textarea' : 'input';
+  const showError =
+    (touched === undefined || touched) && error && (submitCount === undefined || submitCount > 0);
```
carlosbollero, line 46: Couldn't we assign a default value to `touched` and `submitCount` so we can avoid checking for `undefined`?

generators/app/templates/src/app/components/FormInput/styles.module.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.error-text {
+  color: $red;
+  font-size: 12px;
+  height: 20px;
+  opacity: 0;
+
+  &.visible {
+    opacity: 1;
+  }
```
carlosbollero: What about changing this and use `display` or `visibility` attributes instead? Take into account that `display: none;` removes the elem from the layout while `visibility: hidden;` hides it but reserves the space. I think `opacity` would make sense when dealing with a transition from transparent to opaque.

carlosbollero (changes requested):

generators/app/templates/src/app/components/FormInput/index.tsx
briangerez: I think that if you put a value on it, it is not generic.
The idea is to use this behavior regardless of the use of a library for forms.

README.md
briangerez: Actually the suspense document would have to be created 👀

carlosbollero (approved):

PR #124, opened 2020-04-15
made most components optional

damfinkel:
## Summary

- Made Checkbox, RadioGroup and SearchBar optional components. They are selectable when bootstrapping
- Migrated some promises from chains to async/await because of a race condition installing dependencies.

## Screenshots
image

## Trello card

https://trello.com/c/aOPcH3hH/245-sacar-las-librer%C3%ADas-opcionales

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -53,18 +53,17 @@ class GeneratorReact extends Generator {
     }
   }
 
-  configuring() {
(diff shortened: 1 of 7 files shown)


pabloferro (approved):
I don't kown if I agree with having optional components without an easy way to add them after bootstraping the project.
But the PR looks good!

PR #127, opened 2020-04-20
Login functionality

tomiir:
## Summary
Added Login functionality. Login now executes an async request to a mocked service when submitted.
- Migrated AuthService and LocalStorageService to TypeScript. 
- Added a generic login action to be used if needed. 
- Added funcitonality to the submit button, making it execute an async request to a mocked service. The request sets the loading for the screen and on success saves the user on LocalStorage via the LocalStorageService. 

## Screenshots
login-loading

## Trello Card
https://trello.com/c/1pmjH6dF/244-login-v2-funcionalidad

## Depends on 
https://github.com/Wolox/react-bootstrap/pull/116


generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const setCurrentUser = (currentUser: User) => {
+  api.setHeader('Authorization', currentUser.sessionToken);
+  LocalStorageService.setValue(TOKEN_FIELD_NAME, currentUser.sessionToken);
+};
+
+export const getCurrentUser = () => {
+  const currentSessionToken = LocalStorageService.getValue(TOKEN_FIELD_NAME);
+  if (currentSessionToken) {
+    api.setHeader('Authorization', currentSessionToken);
```
damfinkel: I'm not sure if we should do this here. I realize you only execute this function when the app is loaded but we might want to do it elsewhere without setting the header. I think I'd split it, or maybe just take the api part and do that in the component

generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,43 @@
...
+    return true;
+  }
+  return false;
+};
+
+export const removeCurrentUser = () => LocalStorageService.removeValue(TOKEN_FIELD_NAME);
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const login = (credentials: Credentials): Promise<ApiResponse<User, LoginError>> =>
+  // TODO: Implement call to authentication API here
```
damfinkel, line 26: leave a comment with `api.post('login', credentials)`

damfinkel (changes requested):

generators/app/templates/src/services/AuthServices.ts
tomiir: We are actually setting it on the `setCurrentUser` function which is used in the component, so I'll delete it from here. However, it seems logical to me that when we set the current user we set the corresponding header to the api.

generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,37 @@
...
+import LocalStorageService from './LocalStorageService';
+
+const TOKEN_FIELD_NAME = 'sessionToken';
+
+export interface LoginError {
+  message: string;
+}
+
+export const setCurrentUser = (currentUser: User) => {
+  api.setHeader('Authorization', currentUser.sessionToken);
```
damfinkel, line 16: This is fine I guess, but you're not removing the header when you remove user.

damfinkel (changes requested):

generators/app/templates/src/app/screens/Login/styles.module.scss
```diff
@@ -1,20 +1,45 @@
...
 .section-container {
-  width: 100%;
   max-width: 280px;
+  width: 100%;
+}
+
+.error-message {
+  background-color: $red;
+  border-radius: 4px;
+  color: white;
```
SKOLZ: us css variables

generators/app/templates/src/app/screens/Login/styles.module.scss
```diff
@@ -1,20 +1,45 @@
...
-  height: 30px;
   border: solid 1px;
   border-radius: 5px;
+  height: 30px;
+  padding: 5px;
 }
 
 .form-container {
   min-height: 500px;
+  padding-top: 80px;
```
SKOLZ, line 18: I think you need a separate container, if the form will be a box the container should have the padding.

generators/app/templates/src/app/screens/Login/layout.tsx
```diff
@@ -38,16 +45,24 @@ function Login({ onEmailChange, onPasswordChange, onLogin }: Props) {
...
         />
       </div>
       <div className={`column center ${styles.sectionContainer}`}>
-        <button type="submit" className={`full-width m-bottom-1 ${styles.button}`}>
+        <button disabled={loading} type="submit" className={`full-width m-bottom-1 ${styles.button}`}>
           {i18next.t('Login:enter')}
         </button>
+        <span
+          className={`row center middle full-width m-top-1 m-bottom-1
+          ${isError ? styles.errorMessage : styles.hidden}`}
```
SKOLZ: do we want to hide it from css or make it disapear from the dome? if css is the final decision make the class global.

generators/app/templates/src/app/screens/RecoverPassword/styles.module.scss
```diff
@@ -0,0 +1,22 @@
...
+.input {
+  border: solid 1px;
+  border-radius: 5px;
+  height: 30px;
+  padding: 5px;
+}
+
+.form-container {
+  min-height: 500px;
+  padding-top: 80px;
```
SKOLZ: same as before

generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,37 @@
...
+  LocalStorageService.setValue(TOKEN_FIELD_NAME, currentUser.sessionToken);
+};
+
+export const getCurrentUser = () => LocalStorageService.getValue(TOKEN_FIELD_NAME);
+
+export const removeCurrentUser = () => LocalStorageService.removeValue(TOKEN_FIELD_NAME);
```
SKOLZ, line 22: we should add a logout in the home that calls this and calls the api to invalidate the token

generators/app/templates/src/app/screens/Login/layout.tsx
```diff
@@ -3,24 +3,30 @@ import i18next from 'i18next';
...
   return (
-    <form className={`column center full-width m-top-8 ${styles.formContainer}`} onSubmit={onLogin}>
+    <form className={`column center full-width ${styles.formContainer}`} onSubmit={onLogin}>
       <div className="column center m-bottom-3">
-        <h2 className="m-bottom-1">{i18next.t('Login:login')}</h2>
-        <h3>{i18next.t('Login:loginExplanation')}</h3>
+        <h1 className="m-bottom-1">{i18next.t('Login:login')}</h1>
+        <h2>{i18next.t('Login:loginExplanation')}</h2>
       </div>
       <div className={`column m-bottom-2 ${styles.sectionContainer}`}>
```
SKOLZ: I think the form could be the section container and the section container could be the form. that way you'll have a div that wraps everything and a for which contains the fileds.

SKOLZ (changes requested):

generators/app/templates/src/app/screens/Login/layout.tsx
```diff
@@ -38,16 +45,24 @@ function Login({ onEmailChange, onPasswordChange, onLogin }: Props) {
...
       <div className={`column center ${styles.sectionContainer}`}>
-        <button type="submit" className={`full-width m-bottom-1 ${styles.button}`}>
+        <button disabled={loading} type="submit" className={`full-width m-bottom-1 ${styles.button}`}>
           {i18next.t('Login:enter')}
         </button>
+        <span
+          className={`row center middle full-width m-top-1 m-bottom-1
+          ${isError ? styles.errorMessage : styles.hidden}`}
+        >
+          {i18next.t(`Login:${loginError?.errorData?.message}`)}
```
carlosbollero: If `loginError` can be `null` or `undefined`, then an incorrect string would be formed for those cases.
Also, you've already saved that info in `isError`.

carlosbollero (changes requested):

generators/app/templates/src/services/AuthServices.ts
tomiir: will add the counterpart on logout

carlosbollero (approved):

damfinkel (approved):

generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const login = (credentials: Credentials): Promise<ApiResponse<User, LoginError>> =>
+  // TODO: Implement call to authentication API here
+  // api.post('/login', credentials );
+  new Promise(resolve => {
+    setTimeout(() => {
+      resolve({
```
SKOLZ, line 30: there's an existing standarized API that you can assume is being used, talk to Lucas Mujica to get the endpoints

generators/app/templates/src/services/AuthServices.ts
```diff
@@ -0,0 +1,50 @@
...
+        data: { sessionToken: credentials.password === 'asd1' ? 'invalid' : 'token', id: 1234 },
+        problem: null,
+        originalError: null
+      });
+    }, 1000); // eslint-disable-line no-magic-numbers
+  });
+
+
+export const logout = (): Promise<ApiResponse<User, LoginError>> => 
+  new Promise(resolve => {
```
SKOLZ, line 41: same here

SKOLZ (changes requested):

SKOLZ (approved):

SKOLZ (approved):

PR #130, opened 2020-05-08
List example

pabloferro:
## Summary
List example and Paginator component
Pending: make it optional

## Screenshots
list-example


generators/app/templates/src/app/screens/Dashboard/screens/ListExample/reducer.ts
```diff
@@ -0,0 +1,42 @@
+export interface HomeState {
```
damfinkel: If you want to keep this as an example of reducer/state for the screen fine, but change the name so it's called ListExampleState or something like that and use it, otherwise it's an unused file.

damfinkel (changes requested):

generators/app/templates/src/app/screens/Dashboard/screens/ListExample/reducer.ts
pabloferro: Yeah I'll remove this files

generators/app/templates/src/app/screens/Dashboard/screens/ListExample/index.tsx
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable react/no-multi-comp */
```
carlosbollero: Is comment this required?

generators/app/templates/src/app/screens/Dashboard/screens/ListExample/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      ) : (
+        <>
+          {list?.page?.map(item => (
+            <div key={item} className={styles.item}>
+              {item}
+            </div>
+          ))}
+          <Paginator
+            pageCount={list?.totalPages}
+            onPageChange={currentPage => getPage({ page: currentPage.selected + 1 })}
```
carlosbollero, line 28: This just change to the next page, what about the previous?

carlosbollero (changes requested):

damfinkel (approved):

generators/app/templates/src/app/screens/Dashboard/screens/ListExample/index.tsx
pabloferro: Nope, thanks 😄

generators/app/templates/src/app/screens/Dashboard/screens/ListExample/index.tsx
pabloferro: I add one because Paginator starts at page 0, and the service at page 1

carlosbollero (approved):

FrankIglesias (approved):

FrankIglesias:
Check why the CI explodes

mcallegari10 (approved):

pabloferro:
This component was added to the wolox cookbook in https://github.com/Wolox/frontend-cookbook/pull/68

PR #152, opened 2021-01-08
react-i18next migration

alejobh:
## Summary

* Migrated from using `i18next.t` function to use `react-i18next` with hooks
* Added `en` wording
* Changed some component names
* Added jest mock to `react-18next`
* Updated tests and snapshots
* Language change works

## Screenshots

changelang

### Tests
image


generators/app/templates/src/config/api.js
```diff
@@ -25,7 +25,7 @@ export const apiSetup = dispatch => {
     if (response.status === STATUS_CODES.unauthorized) {
       /*
        * TODO: These callbacks should only be called if no other callback was asigned for the response.
-       * - dispatch(alertActions.error(i18next.t('apiErrors:expired')));
+       * - dispatch(alertActions.error(t('apiErrors:expired')));
```
pabloferro: Where do we get `t` from here? 😅

generators/app/templates/src/config/i18n.js
```diff
@@ -1,12 +1,18 @@
...
+import i18next from 'i18next';
+import { initReactI18next } from 'react-i18next';
 
 function requireAll(requireContext) {
   return requireContext.keys().map(requireContext);
 }
 
-init({
+i18next.use(initReactI18next).init({
+  fallbackLng: 'en',
```
pabloferro: Why is the fallback 'en', can we leave it in 'es'?

generators/app/templates/src/screens/Registration/index.tsx
```diff
@@ -14,34 +15,42 @@ import styles from './styles.module.scss';
...
+    confirmPassword: requiredValidation
+  };
 };
 
-function RegistrationContainer() {
+function Registration() {
   const { register, handleSubmit, errors, watch } = useForm();
   const dispatch = useDispatch();
   const history = useHistory();
+  const { t } = useTranslation(['Registration', 'RecoverPassword']);
```
pabloferro: Why do you need `RecoverPassword` here? I'd keep the screens independant.

pabloferro (changes requested):
Awesome work! I just have some questions.

generators/app/templates/src/config/api.js
alejobh: should be sent to `apiSetup` on a higher lvl component that uses `const { t } = useTranslate('apiErrors')' // or [..., 'apiErrors'])`

generators/app/templates/src/config/i18n.js
alejobh: i can change it but makes no sense because the default language is `es`

generators/app/templates/src/screens/Registration/index.tsx
alejobh: i kept the old translations but i can move those translations here

generators/app/templates/src/config/i18n.js
pabloferro: So why do we need a default and fallback? I'm assuming that most projects will only have spanish, so I'd change it to `es`

generators/app/templates/src/screens/Registration/index.tsx
pabloferro: I'd only keep `Registration`

generators/app/templates/src/config/api.js
damfinkel: I don't think we should use dispatch from here. I'd delete this and if eventually someone wants to do this, they can pass a callback instead, so this setup doesn't depend on the way the error is called/saved

generators/app/templates/src/config/api.js
```diff
@@ -20,12 +20,12 @@ const api = create({
 });
 
 // eslint-disable-next-line no-unused-vars, prettier/prettier, @typescript-eslint/no-unused-vars
-export const apiSetup = dispatch => {
+export const apiSetup = callback => {
```
damfinkel: instead of callback rename it to `errorCallback` or `onError` in case we want other callbacks for other cases.

damfinkel (changes requested):

generators/app/templates/src/config/api.js
alejobh: what if there is a strange case for someone that needs a monitor for an ok response or something like that?

pabloferro: But this callback is called only on 401. I'd rename it to `unauthorizedCallback`.
If you need multiple monitors maybe it would be better to call addMonitor directly in the app.

pabloferro (approved):

damfinkel (approved):

PR #153, opened 2021-01-12
route testing

damfinkel:
## Summary

Tested Routes and RouteItem

## Screenshots

image
image


alejobh (approved):

pabloferro (approved):
Looks good. I have one question, should we add a test with each new route when working in a project?

alejobh:
> Looks good. I have one question, should we add a test with each new route when working in a project?

I think no, he is testing each route that we have in the bootstrap, but in a real scenario, those tests should be on the screen level

PR #158, opened 2021-03-03
Add component Generator

pabloferro:
## Summary

Add component Generator

What do you think about the templates?

I'm not sure if they have to much code. It serves as documentation for our standards, but you'll probably need to delete some code every time you generate some component.

Also everyone could modify the templates in their projects when needed.


generators/app/templates/templates/component/index.test.tsx.hbs
```diff
@@ -0,0 +1,18 @@
...
+  const { container } = render(<{{name}} someProp={1} />);
+  expect(container.firstChild).toMatchSnapshot();
+});
+
+test('renders correctly with the optionalProp', () => {
+  const { container } = render(<{{name}} someProp={1} someOptionalProp={2} />);
+  expect(container.firstChild).toMatchSnapshot();
+});
+
+// You can test simple components with snapshots but if
```
damfinkel: I'd move this comment above the first test

damfinkel (approved):
I love it! Pretty simple and useful

generators/app/templates/templates/component/index.test.tsx.hbs
alejobh: Same as @damfinkel and the comment could be `You can use snapshot tests for simple components, but if the component has any interaction or conditional logic you should add tests for those.`

alejobh (approved):

PR #159, opened 2021-03-03
changed linter packages, updated dependencies and removed files added…

damfinkel:
… to cookbook

## Summary

- Added new linter packages
- Fixed linter warnings related to parenthesis (prettier changed the default behaviour un v2)
- Removed useRequest, utils, Login and Register whichwere already moved to the cookbook


generators/app/steps/CustomizableSteps/constants.js
```diff
@@ -1,12 +1,10 @@
 const COMPONENTS_PATH = 'src/components';
 const SCREENS_PATH = 'src/screens';
 const CONFIG_PATH = 'src/config';
-const UTILS_PATH = 'src/utils';
 const SERVICES_PATH = 'src/services';
 const CONSTANTS_PATH = 'src/constants';
 const STYLES_PATH = 'src/scss';
 const DOCS_README_PATH = 'docs';
-const HOOKS_PATH = 'src/hooks';
```
pabloferro: I'd preserve these folders with a README.md. What do you think?

generators/app/templates/src/screens/Dashboard/screens/Home/index.tsx
```diff
@@ -13,24 +9,17 @@ import { reducer, INITIAL_STATE, actionCreators } from './reducer';
...
   useEffect(() => {
     dispatch(actionCreators.setFoo('React'));
   }, [dispatch]);
 
   const handleLogout = () => {
-    userDispatch(authActions.logout());
-    logoutRequest(null);
+    // Perform logout request and on success add the following:
+    // userDispatch(authActions.resetUser());
+    // removeCurrentUserToken();
```
pabloferro: Why did you remove this dispatch?

pabloferro (changes requested):
Looks good! Just two comments.

generators/app/templates/src/screens/Dashboard/screens/Home/index.tsx
pabloferro: Also check conflicts

damfinkel: I didn't remove it, I purposedly left them commented because I thought they should be in the logout success callback. But now that I think of it, we should perform these two actions even if logout fails I suppose. So I'll uncomment them

alejobh (approved):

generators/app/templates/src/components/Routes/constants.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const ROUTES = [
+  // Leaving this as an example for then the Login screen exists
+  {
+    exact: false,
+    path: PATHS.login,
+    component: () => <span>login</span>,
+    title: 'Routes:loginTitle',
+    description: 'Routes:loginDescription',
+    redirectTo: (user: User | null) => (user ? MAIN_PRIVATE_PATH : undefined)
+  },
```
pabloferro, line 24: I don't think we need this example if we have a generator, maybe we can add a comment here about the generator instead.

generators/app/templates/src/utils/README.md
```diff
@@ -0,0 +1,3 @@
+### Global utils files
+
+Add global utils functions and files in this folder. You can check for possible global utils from our [cookbook](https://cookbook.wolox-fearmy.now.sh/recipes/utils)
```
pabloferro, line 3: Nice

pabloferro (approved):

PR #166, opened 2021-05-12
Test coverage, Home and lint updates 

alejobh:
## Summary
* Modified `Home` component to use i18n and use the feature to change language
* Added missing tests
* Updated some tests
* Modified eslint versions to correctly lint files

## Screenshots
image


generators/app/templates/README.md
```diff
@@ -16,6 +16,17 @@ You can use `npm run generate` to add different components to your app like:
 * Components
 * Screens
 
+## Using react-i18next
+
+Add your resource boundle translations on component level and import that file on top of that component to include translations.
```
pabloferro:
```suggestion
Add your resource bundle translations on the component level and import that file in the component to include translations.
```

generators/app/templates/src/components/App/index.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+jest.mock(
+  'components/Routes',
+  () =>
+    function Routes() {
+      return <span>Routes</span>;
+    }
+);
+
+describe('App component', () => {
+  test('Renders without errors', () => {
```
pabloferro:
```suggestion
  test('renders without errors', () => {
```

generators/app/templates/src/components/App/index.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+jest.mock(
+  'components/Routes',
+  () =>
+    function Routes() {
+      return <span>Routes</span>;
+    }
+);
+
+describe('App component', () => {
```
pabloferro: Maybe remove this describe?

generators/app/templates/src/components/FormInput/index.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { render, screen } from '@testing-library/react';
+
+import FormInput from '.';
+
+const name = 'aName';
+const label = 'aLabel';
+
+describe('FormInput component', () => {
```
pabloferro: Maybe remove this describe?

generators/app/templates/src/config/i18n.js
```diff
@@ -1,19 +1,12 @@
...
-function requireAll(requireContext) {
-  return requireContext.keys().map(requireContext);
-}
-
 i18next.use(initReactI18next).init({
   lng: 'es',
-  // uncomment next line if you have multilang in your application and you have a fallback language
-  // fallbackLng: 'en',
+  supportedLngs: ['es', 'en'],
+  // Comment next line if your application does not have multilang support
```
pabloferro:
```suggestion
  // Remove next line if your application does not have multilang support
```
We shouldn't promote commented code 😅

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    jest.clearAllMocks();
+  });
+
+  test('logouts when clicks logout button', async () => {
+    const logout = jest.fn();
+    const resetUser: () => ResetUser = jest.fn(() => ({ type: ActionTypes.RESET_USER }));
+    const removeCurrentUserToken = jest.fn();
+    jest.spyOn(AuthService, 'logout').mockImplementationOnce(logout);
+    jest.spyOn(authActions, 'resetUser').mockImplementationOnce(resetUser);
+    jest.spyOn(AuthService, 'removeCurrentUserToken').mockImplementationOnce(removeCurrentUserToken);
```
pabloferro: I think all these mocks are signs of something being wrong.
Maybe `removeCurrentUserToken` should be called inside `logout` since both are part of the `AuthService`?

Also why do you need to know the implementation of `resetUser`? Can you just mock `const resetUser = jest.fn();`

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
```diff
@@ -0,0 +1,72 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { actionCreators as authActions, ActionTypes, ResetUser } from 'contexts/UserContext/reducer';
+import { withContextProvider as withUserContext } from 'contexts/UserContext';
+import * as AuthService from 'services/AuthService';
+
+import { withContextProvider as withTechContext } from './context';
+
+import Home from '.';
```
pabloferro, line 6:
```suggestion
import Home from '.';

const HomeWithUserContext = withUserContext(Home);
```

You can define the const here and use it in the tests.

I'm not sure if we should use the `withUserContext` since this a HomeTest. You could mock the context.

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  test('Calls change language from en to es', async () => {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const RenderHOC = withTechContext(Home);
+    render(<RenderHOC />);
+    // Calls once for changing mocked en lang to es
+    userEvent.click(screen.getByRole('button', { name: /Home:changeLang/ }));
+    await waitFor(() => expect(screen.queryByText(/Home:loggedIn/)).toBeInTheDocument());
+  });
```
pabloferro: We are expecting the same thing in both tests, do we have a way to tell if it's in spanish or englidh?

Maybe mocking i18n? If not this test doesn't work.

pabloferro (changes requested):
Awesome work! I left some comments.

generators/app/tasks/fileCreators.js
```diff
@@ -5,7 +5,14 @@ const getPackageJsonAttributes = (projectName, projectVersion, repoUrl) => ({
   title: projectName,
   version: projectVersion,
   jest: {
-    collectCoverageFrom: ['src/**/*.{js,jsx,ts,tsx}'],
+    resetMocks: true,
+    collectCoverageFrom: [
```
damfinkel, line 9: Should we ignore all i18n.ts? Since we mock it, tests never cover them

generators/app/templates/src/screens/Dashboard/screens/Home/context/index.test.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable max-nested-callbacks */
+import { renderHook } from '@testing-library/react-hooks';
+import { useContext } from 'react';
+
+import { INITIAL_STATE } from './reducer';
+
+import { useSelector, useDispatch, StateContext, DispatchContext } from '.';
+
+describe('When wrapping a component with the context', () => {
```
damfinkel: unnecessary dscribe

generators/app/templates/src/screens/Dashboard/screens/Home/context/reducer.test.ts
```diff
@@ -0,0 +1,22 @@
+import { actionCreators, reducer } from './reducer';
+
+const EMPTY_STATE = { tech: '' };
+
+test('setTech action sets the new tech', () => {
+  const newtech = 'Angular';
```
damfinkel: newTech

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    jest.spyOn(AuthService, 'removeCurrentUserToken').mockImplementationOnce(removeCurrentUserToken);
+
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const RenderHOC = withUserContext(Home);
+    render(<RenderHOC />);
+
+    const logoutButton = screen.getByRole('button', { name: /Home:logout/ });
+    userEvent.click(logoutButton);
+
+    await waitFor(() => expect(logout).toHaveBeenCalled());
```
damfinkel, line 26: is this waitfor necessary? If it is, shouldn't we waitFor the click, which I guess is the code that could take some time? If you keep the waitFor and move it to userEvent.click, change the others so all waitFors do the same

damfinkel (changes requested):

generators/app/tasks/fileCreators.js
alejobh: Since we cannot cover that, the only way to cover is to create a `i18n.test.tsx` with the `addResources` mocked, but it adds nothing, is a waste of time to always add the same code mocking something that should be tested by `i18next` lib. That's why we are ignoring them

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
alejobh: the click is synchronous but will trigger some changes in the component, that changes are waited by `waitFor` that is why is necessary.

generators/app/templates/src/screens/Dashboard/screens/Home/index.test.tsx
alejobh: services are unit tested, because the component uses those services, they dont need to be tested again, that is why i dont care about the implementation, the component only needs to know that the function is actually called, and all of those services are called in the logout handle click. So the easiest way to retrieve coverage from it is to mock the implementation and check if it was actually called. Otherside, if we dont mock that services or actions, the code will work as the implementation were done. So we expect any graphic change to test but the logout action does not have any, the only thing is the redirection.

generators/app/tasks/fileCreators.js
alejobh: Adding to the previous comment: I forgot we are not ignoring i18n files right now, we changed the way we include translations by adding an `import './i18n';` in the component that creates a i18n file. first to add coverage to them. Second to avoid adding test to them.

pabloferro (approved):

generators/app/tasks/fileCreators.js
damfinkel: I know we weren't ignoring them, I was suggesting to do so. I think importing in each component the i18n file is a little uncomfortable if we can just fix this by ignoring the i18n files when calculating the coverage

damfinkel (approved):

pabloferro (approved):

PR #170, opened 2021-06-28
node engine version set to 14.x

alejobh:
## Summary

* node engine version set to `14.x` in `bootstrap` and `bootstraped project`
Tested in `14.0.0`, `14.15.1`, `14.17.1` lastest LTS version


damfinkel (commented):
We need to make the followinf changes as well:

1) Change Dockerfile to use node:14-alpine3.14 (both this project's and the generated one's)
1) Fix versions of these two packages: eslint-plugin-prettier (3.3.1) and prettier (2.2.1), because there's a clash between prettier and eslint from that version on with which we already had some problems in other projects

damfinkel (approved):

PR #178, opened 2022-02-24
Updated CRA to v5.0.0 and updated eslint

MMarussi:
## Summary

- Updated CRA and react scripts from 4.0.1 to 5.0.0
- Added eslint v8.9.0 to dev dependencies
- Updates some eslint plugins in order to be compatible with new eslint version
- Updated tests to comply with new testling-library-eslint rules

## Known issues

- New tests throw the `state changes not wrapped in act()` warning
- It would seem that CRA v5 dropped the git init command, so if the user chooses to not use a repo via url it will create it but it wont be configured as a git repo, so the method `configGitNoRepo()` will fail and wont commit any file


generators/app/templates/src/components/Routes/index.test.tsx
```diff
@@ -24,7 +24,7 @@ describe('when there is a user', () => {
   test('shows Home screen when being on the home path', async () => {
     window.history.pushState({}, '', PATHS.home);
     render(<WrappedRoutes userState={userState} />);
-    await waitFor(() => expect(screen.getByText(/Home:loggedIn/)).toBeInTheDocument());
+    await screen.findByText(/Home:loggedIn/);
```
pabloferro: We need to add a `expect` here.

generators/app/templates/src/components/Routes/index.test.tsx
```diff
@@ -34,6 +34,6 @@ describe('when there is no user', () => {
   test('redirects to Login screen when on the home path', async () => {
     window.history.pushState({}, '', PATHS.home);
     render(<WrappedRoutes userState={userState} />);
-    await waitFor(() => expect(screen.getByText(/login/)).toBeInTheDocument());
+    await screen.findByText(/login/);
```
pabloferro: We need to add a `expect` here.

pabloferro (approved):

pabloferro (approved):

